PR #8, opened 2023-11-09
Add basic repo configuration

rainest:
Add a basic README, copy basic tooling configuration from deck main, and copy LICENSE from deck main.

Configuration edited to remove filepath-specific configuration that I don't expect will migrate here.

LICENSE date updated, but otherwise unchanged. Unsure if we need to get OSS approval for this because it's just moving existing OSS code to a new home. Working with the assumption that we do not.

Removes the template repo pre-commit and sync actions. This does not appear properly configured for Golang, as it includes checks that [prohibit the use of tabs](https://github.com/Kong/go-database-reconciler/actions/runs/6806112060/job/18506829076?pr=9). IDK what the status of that repo template is, but it looks like it'd need changes if we're expected to use it for Golang projects.


.golangci.yml
```diff
@@ -0,0 +1,79 @@
...
+  - nilerr
+  - nolintlint
+  - predeclared
+  - revive
+  - stylecheck
+  - unconvert
+  - unparam
+  - wastedassign
+  - errorlint
+issues:
```
pmalek, line 30:
```suggestion
issues:
  max-same-issues: 0
```

This will make us immediately get the full occurrence list of (the same) issues instead of capping it to 3.

pmalek (approved):

PR #9, opened 2023-11-09
Migrate deck@v1.29.2 library code into this repo

rainest:
Copy the `konnect`, `cprint`, `crud`, `diff`, `dump`, `file`, `scripts`, `state`, `types`, and `utils` packages from https://github.com/Kong/deck/releases/tag/v1.29.2 (also the current tip of main) into this repository's `pkg` dir.

Update references to those packages from their deck location to this repository.

Run:

```
go mod init "github.com/kong/go-database-reconciler"
go mod tidy
```   

Add replace directive from deck go.mod.
    
Run:

```    
go get github.com/alecthomas/jsonschema@v0.0.0-20191017121752-4bb6e3fae4f2
go mod tidy
```
    
This library is apparently rather out of date in deck and has had breaking changes since it was added. After this downgrade, `go vet  ./...` no longer shows errors.

Issue #7, opened 2023-11-09
Determine test plan


rainest:
Existing deck integration tests generally rely on the CLI. This repo will not have that CLI readily available, and will need some changes to its testing strategy.

Offhand, I'd expect the simplest option is have the test actions here check out a copy of the deck repo main, update its go.mod to use the PR head commit as the current version of this library, and run deck's usual test suite. This approach will probably hit chicken-egg problems where coverage for this repo relies on features that aren't yet merged into deck main. Unsure how to best handle that, but we can maybe add a variant of tests that lets you choose the downstream branch and still satisfies the same Actions status.

rainest:
https://github.com/Kong/go-database-reconciler/pull/10 is the most basic option for this. It reuses existing tests by importing deck as a test dependency here.

This generally works, though creates a (test only) dependency chain that's a bit silly. To avoid it we'd need to implement some sort of alternate to the client code, that invokes the sync/diff functions here without either of the downstream consumers (deck and KIC) of this library.

PR #10, opened 2023-11-14
Add CI tests

rainest:
Will it test? That is the question.

Essentially copying the test CI and code from deck after https://github.com/Kong/deck/pull/1109. Will need to figure out what to snip out later.


GGabriele:
@rainest assuming (also based on the commits) this PR is basically [this other PR](https://github.com/Kong/go-database-reconciler/pull/9) plus tests, can we use `chore/lift-and-shift` as base branch for this PR in order to help with diffs?

rainest:
Sure.

.ci/check.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash -ex
+
+diff -u <(echo -n) <(gofmt -d -s .)
```
pmalek, line 3: `gofmt` when invoked like this will set the exit code so if `set -e` is set (I'm not sure if putting `-e` in shebang makes it work like that) then it can be simply made:

```suggestion
gofmt -d -s .
```

.ci/setup_kong.sh
```diff
@@ -0,0 +1,65 @@
...
+  -e "KONG_PROXY_ERROR_LOG=/dev/stderr" \
+  -e "KONG_ADMIN_ERROR_LOG=/dev/stderr" \
+  -e "KONG_ADMIN_LISTEN=0.0.0.0:8001, 0.0.0.0:8444 ssl" \
+  -p 8000:8000 \
+  -p 8443:8443 \
+  -p 127.0.0.1:8001:8001 \
+  -p 127.0.0.1:8444:8444 \
+  $KONG_IMAGE
+
+waitContainer "Kong" 8001 0.2
```
pmalek, line 65:
```suggestion
waitContainer "Kong" 8001 0.2

```

tests/integration/test_utils.go
```diff
@@ -0,0 +1,344 @@
...
+	"context"
+	"io"
+	"os"
+	"testing"
+
+	"github.com/acarl005/stripansi"
+	"github.com/fatih/color"
+	"github.com/google/go-cmp/cmp"
+	"github.com/google/go-cmp/cmp/cmpopts"
+	"github.com/kong/deck/cmd"
```
pmalek, line 14: Hm as [noted in the linked issue](https://github.com/Kong/go-database-reconciler/issues/7#issuecomment-1812956823) this might not be breaking the build but is rather undesirable 🤔 

Couldn't we leave whatever's using the `cmd` in deck repo and test here without resorting to deck commands? This way we'd break up the dependency. Not sure how feasible that is given that most of the code here relies on issuing requests against Kong (or Konnect) via the cmd

rainest: As designed the integration tests are very much designed around using the CLI commands as they would be used at the CLI. We could build a pseudo-CLI to call the various package functions in sequence here, but in practice that code would be almost identical to the CLI minus cobra and such. IDK how much we'd need the separation in form if not function to meet the goals of the separation--that's more a question for @Tieske probably.

My hope is that after the initial separation we do continue with further refactors to make the packages and their APIs here better designed as modular components, and in the course of those refactors we write new integration tests that slowly replace the original deck ones.

Is there, alternately, a way to automatically run an external repo's tests against a PR branch as part of that PR's checks? We have good reason to run deck and KIC tests against proposed library changes regardless. Being able to do that without actually importing their packages here feels less backwards, for lack of a better word.

rainest:
Re-added Codecov now that I've realized this doesn't require an org admin to configure and added a token.

In deck (and KIC) we weirdly have the CODECOV_TOKEN in a `configure CI` environment, but we _don't_ use this environment in the workflow configuration (which apparently doesn't cause any issues). I'm not sure what's up with that. I've just placed it in the main repo secrets for now.

tests/integration/test_utils.go
pmalek: If you wanted to run deck and/or KIC tests with changes in this repo then simply checking those out via something like https://github.com/Kong/kubernetes-ingress-controller/blob/44ef693b3cae49a55475cde4097af28739f36918/.github/workflows/release_docs.yaml#L26-L31 and then:

```
go mod edit -replace=github.com/kong/go-database-reconciler=../go-database-reconciler
go mod tidy
```

(assuming checkout to the same parent dir)

---

As for the refactors: 👍. It would be highly desirable to remove that (semi-)circular dependency between this repo and deck.

We'd basically need to peel off one layer (cmd invocation) and use the actual library code (but that's just my speculation). deck tests should be left as is anyway to utilize this lib in its integration suite.

GGabriele (commented):
minor cleanup comments

Makefile
```diff
@@ -0,0 +1,52 @@
...
+
+.PHONY: test
+test:
+	go test -race -count=1 ./...
+
+.PHONY: lint
+lint:
+	golangci-lint run -v ./...
+
+.PHONY: build
```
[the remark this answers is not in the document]
GGabriele: Actually, I see that the `scripts` folder is present in the base branch from [the other PR](https://github.com/Kong/go-database-reconciler/pull/9/files), but when I pulled `chore/test` I couldn't find it. Is this missing a rebase?

.github/workflows/integration.yaml
```diff
@@ -0,0 +1,51 @@
+name: Integration Test
```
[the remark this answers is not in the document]
rainest: The collection of repo steps required to grant it access isn't exactly documented, so I had to go digging around for the credentials. It's there now.

.github/workflows/integration.yaml
```diff
@@ -0,0 +1,51 @@
...
+        - 'kong:2.3.3'
+        - 'kong:2.4.1'
+        - 'kong:2.5.1'
+        - 'kong:2.6.0'
+        - 'kong:2.7'
+        - 'kong:2.8'
+        - 'kong:3.0'
+        - 'kong:3.1'
+        - 'kong:3.2'
+        - 'kong:3.3'
```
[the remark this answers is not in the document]
rainest: The other fix I pulled in here isn't what un-breaks these, so this is deferred to https://github.com/Kong/go-database-reconciler/issues/15

GGabriele (approved):

PR #25, opened 2023-12-01
chore: add a test aggregation workflow

rainest:
Re https://github.com/Kong/go-database-reconciler/pull/21#issuecomment-1836186240 why things merged without passing tests, well, we simply have no required tests! We have a require status checks branch protection rule with no required status checks configured, same as deck did. I'm not sure if anything was preventing merges with failures in the past.

This cannibalizes a similar [KIC workflow](https://github.com/Kong/kubernetes-ingress-controller/blob/main/.github/workflows/checks.yaml) that aggregates required test workflows into a single `passed` status. Note that we still need to add this to the branch protection rules after for it to have effect.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -5,8 +5,7 @@ updates:
   schedule:
     interval: daily
   ignore:
-  - dependency-name: "github.com/fatih/color "
-    versions: ">=1.16.0"
+  - dependency-name: "github.com/fatih/color"
 - package-ecosystem: github-actions
   directory: /
   schedule:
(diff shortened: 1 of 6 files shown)


Tieske (approved):

rainest:
Now includes (the third or fourth) downgrade attempt also, since dependabot [is too fast](https://github.com/Kong/go-database-reconciler/pull/28) and jumps in to help before tests here could run again and pass the merge requirements :|

Tieske (approved):

PR #31, opened 2023-12-23
Feat/default add look up tags REBASED

AntoineJac:
Linked to the PR:
https://github.com/Kong/deck/pull/1124

Description:
Please see Jira: https://konghq.atlassian.net/browse/APIOPS-100

This branch will add a new feature to pass a new lookup_tags in the _info section. It will allow the local state to retrieve the the corresponding entities in Kong. 

```
_info: 
  default_lookup_tags:
    consumers:
    - test
    - test3
    - test2
```


teb510:
@GGabriele could you review this guy please? then we can merge the 'global consumer' PR after this one.

pkg/dump/dump.go
```diff
@@ -28,6 +28,10 @@ type Config struct {
 	// tags.
 	SelectorTags []string
 
+	// LookUpSelectorTagsConsumers can be used to lookup entities using other
```
GGabriele, line 31: Can you please expand on this? As it is, it's not immediate how this can be used and when.

GGabriele (commented):
@AntoineJac I added a small comment, but other than that this LGTM!

pkg/dump/dump.go
AntoineJac: @GGabriele for instance CMA is creating consumers using a pipeline 1 with tag "consumers_global". 
They have a pipeline 2 for services (tag: serviceName) that could for instance set up "rate-limting" plugins on different consumers, as these consumers are not part of the sync for this pipeline it will fail.

Using a LookUpSelectorTagsConsumers you can specify deck to check if the consumers exist on a different tag

Please let me know if you have other questions

GGabriele (approved):

PR #55, opened 2024-02-07
fix: correctly handle route schema defaults with Konnect

GGabriele:
Kong Gateway supports different schemas for different router versions. On the other hand, Konnect can support only one schema including all fields from 'traditional' and 'expressions' router schemas. This may be problematic when it comes to defaults injection, because the defaults for the 'traditiona' router schema can be wrongly injected into the 'expressions' route configuration.

Here we make sure that only the fields that are supported for a given router version are set in the route configuration.

### Summary

SUMMARY_GOES_HERE

### Full changelog

* [Implement ...]
* [Fix ...]

### Issues resolved

Fix #XXX

### Documentation

### Testing

- [ ] Unit tests
- [ ] E2E tests

### DIFF
--- a/pkg/file/builder.go
+++ b/pkg/file/builder.go
@@ -1110,9 +1110,35 @@ func (b *stateBuilder) ingestRoute(r FRoute) error {
 		return err
 	}
 	r.Route.StripPath = stripPath
+
+	hasExpression := r.Route.Expression != nil
+	hasRegexPriority := r.Route.RegexPriority != nil
+	hasPathHandling := r.Route.PathHandling != nil
+
 	b.defaulter.MustSet(&r.Route)
 	if route != nil {
 		r.Route.CreatedAt = route.CreatedAt
+	}
+
+	// Kong Gateway supports different schemas for different router versions.
+	// On the other hand, Konnect can support only one schema including all
+	// fields from 'traditional' and 'expressions' router schemas.
+	// This may be problematic when it comes to defaults injection, because
+	// the defaults for the 'traditiona' router schema can be wrongly injected
+	// into the 'expressions' route configuration.
+	//
+	// Here we make sure that only the fields that are supported for a given
+	// router version are set in the route configuration.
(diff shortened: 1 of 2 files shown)


czeslavo (approved):
Thanks for that @GGabriele! 🎖️  I verified [it works as expected in KIC](https://github.com/Kong/kubernetes-ingress-controller/pull/5581).

pkg/file/builder.go
```diff
@@ -1110,11 +1110,37 @@ func (b *stateBuilder) ingestRoute(r FRoute) error {
...
 	b.defaulter.MustSet(&r.Route)
 	if route != nil {
 		r.Route.CreatedAt = route.CreatedAt
 	}
 
+	// Kong Gateway supports different schemas for different router versions.
+	// On the other hand, Konnect can support only one schema including all
+	// fields from 'traditional' and 'expressions' router schemas.
+	// This may be problematic when it comes to defaults injection, because
+	// the defaults for the 'traditiona' router schema can be wrongly injected
```
pmalek, line 1127: `traditiona` -> `traditional` 🙃

pkg/file/builder_test.go
```diff
@@ -2755,3 +2755,93 @@ func Test_getStripPathBasedOnProtocols(t *testing.T) {
...
+		t.Run(tt.name, func(t *testing.T) {
+			ctx := context.Background()
+			b := &stateBuilder{
+				currentState: tt.fields.currentState,
+				isKonnect:    true,
+			}
+			b.rawState = &utils.KongRawState{}
+			d, _ := utils.GetDefaulter(ctx, defaulterTestOpts)
+			b.defaulter = d
+			b.intermediate, _ = state.NewKongState()
```
pmalek, line 2840: Too late to the party but any particular reason not to check the errors here?

GGabriele: could be done, but not meaningful for the specific test...copy&pasta mistake

PR #61, opened 2024-02-21
add licenses in kong state and sync diff

### LINKED ISSUE #60: Add support of Licenses
Currently go-database-reconciler does not support CRUD of licenses. This makes KIC fail to update licenses when Kong gateway is DB backed. We need to add support of licenses in GDR.

randmonkey:
### Summary

SUMMARY_GOES_HERE

### Full changelog

* Add support for licenses in Kong state.

### Issues resolved

Fix #60
### Documentation

### Testing

- [ ] Unit tests
- [ ] E2E tests


pkg/diff/diff.go
```diff
@@ -406,6 +417,11 @@ func (sc *Syncer) eventLoop(ctx context.Context, d Do) error {
 }
 
 func (sc *Syncer) handleEvent(ctx context.Context, d Do, event crud.Event) error {
+	// Skip licenses if includeLicense is disabled.
+	// TODO: use a clearer or more general way to ignore licenses when includeLicenses=false?
+	if !sc.includeLicenses && event.Kind == KindLicense {
+		return nil
+	}
```
rainest, line 424: What actually happens if this is absent?

Convention (sorta, there's only one other skippable entity) is to skip [them in the state builder](https://github.com/Kong/go-database-reconciler/blob/fa395292411a55869e7e742f27f6e232eb492984/pkg/file/builder.go#L86-L88), which I think should be equivalent to skipping events (rather, it'd not generate them in the first place).

The earlier proposal to add full license entity support [only included the syncer skip](https://github.com/Kong/deck/pull/882/files#diff-61c9587f2d7a5abbf52f69e0983b087ec5309c08721a20bc4155f57e1119fb6eR267-R276), though I don't recall why it didn't have a builder skip similar to CA certificates--it may well have just been something I should have had, but overlooked.

randmonkey: If we do not skip the event here, the event will be called by the `Do` func in `d(event)`. I will see that if skip them in the builder works.

.ci/clean_kong.sh
```diff
@@ -0,0 +1,29 @@
+#!/bin/bash
+
+set -e
+
+NETWORK_NAME=deck-test
+
+PG_CONTAINER_NAME=pg
+GATEWAY_CONTAINER_NAME=kong
+
+if [[ $(docker ps -a | grep ${GATEWAY_CONTAINER_NAME}) != "" ]]; then
```
pmalek, line 10:
```suggestion
if [[ $(docker ps -a --filter Name=${GATEWAY_CONTAINER_NAME}) != "" ]]; then
```

.ci/clean_kong.sh
```diff
@@ -0,0 +1,29 @@
...
+    echo "remove container ${GATEWAY_CONTAINER_NAME}"
+    docker rm -f  ${GATEWAY_CONTAINER_NAME}
+else
+    echo "container ${GATEWAY_CONTAINER_NAME} not found, skip removing"
+fi
+
+if [[ $(docker ps -a | grep ${PG_CONTAINER_NAME}) != "" ]]; then
```
pmalek, line 17:
```suggestion
if [[ $(docker ps -a --filter Name=${PG_CONTAINER_NAME}) != "" ]]; then
```

.ci/clean_kong.sh
```diff
@@ -0,0 +1,29 @@
...
+    echo "remove container ${PG_CONTAINER_NAME}"
+    docker rm -f  ${PG_CONTAINER_NAME}
+else
+    echo "container ${PG_CONTAINER_NAME} not found, skip removing"
+fi
+
+if [[ $(docker network ls | grep $NETWORK_NAME) != "" ]]; then
```
pmalek, line 24:
```suggestion
if [[ $(docker network ls --filter Name=${NETWORK_NAME}) != "" ]]; then
```

pkg/diff/diff.go
```diff
@@ -47,6 +47,8 @@ type EntityChanges struct {
 
 var errEnqueueFailed = errors.New("failed to queue event")
 
+const KindLicense = "license"
+
```
pmalek, line 51: I might be missing something but this looks unused.

```suggestion
```

pkg/dump/dump.go
```diff
@@ -907,6 +921,34 @@ func GetAllRBACREndpointPermissions(ctx context.Context,
...
+// GetAllLicenses queries Kong for all the Licenses using client.
+func GetAllLicenses(
+	ctx context.Context, client *kong.Client, tags []string,
+) ([]*kong.License, error) {
+	var licenses []*kong.License
+	opt := newOpt(tags)
+
+	for {
+		s, nextopt, err := client.Licenses.List(ctx, opt)
+		if kong.IsNotFoundErr(err) || kong.IsForbiddenErr(err) {
```
pmalek, line 933: We don't return an error when we get a forbidden status?

pkg/file/builder.go
```diff
@@ -938,6 +939,9 @@ func (b *stateBuilder) routes() {
 func (b *stateBuilder) enterprise() {
 	b.rbacRoles()
 	b.vaults()
+	if b.includeLicenses && !b.isKonnect {
```
pmalek, line 942: But be worth adding comments and explaining why we don't fetch license when we're running against Konnect.

pkg/state/types.go
```diff
@@ -1602,3 +1602,41 @@ func (v1 *Vault) EqualWithOpts(v2 *Vault, ignoreID, ignoreTS bool) bool {
...
+	}
+	if ignoreTS {
+		l1Copy.CreatedAt = nil
+		l2Copy.CreatedAt = nil
+
+		l1Copy.UpdatedAt = nil
+		l2Copy.UpdatedAt = nil
+	}
+	return reflect.DeepEqual(l1Copy, l2Copy)
+}
```
pmalek, line 1642: The `License` type and all the methods on it are missing godocs. The other types and functions in this file do have those so I'd suggest we are consistent and add those as well here.

pkg/types/license.go
```diff
@@ -0,0 +1,177 @@
...
+
+// Delete deletes a License in Kong.
+// The arg should be of type crud.Event, containing the license to be deleted,
+// else the function will panic.
+// It returns a the deleted *state.License.
+func (s *licenseCRUD) Delete(ctx context.Context, arg ...crud.Arg) (crud.Arg, error) {
+	if s.isKonnect {
+		return nil, nil
+	}
+	event := crud.EventFromArg(arg[0])
```
pmalek, line 53: Can `arg` be not provided? This will `panic` in that case. WDYT about checking `len(arg)`?

Ditto for all the other use cases above.

randmonkey: This is just the same as the CRUD functions for other types. I think it is OK for checking `len(arg)` and returning errors but this needs some discussion about how to change the existing CRUD functions for the other types.

.ci/clean_kong.sh
randmonkey: done.

pkg/diff/diff.go
randmonkey: removed.

pkg/file/builder.go
randmonkey: Added comments.

pkg/state/types.go
randmonkey: done.

pkg/types/license.go
randmonkey: Added errors for the case.

pkg/dump/dump.go
randmonkey: After tests, Kong does not return 403 when accessing `GET /license` endpoint. So Forbidden should be an error. Updated.

GGabriele:
From a product perspective, LGTM. 
I'll leave the code review to Patryk and Travis who already went through this.

pkg/state/license.go
```diff
@@ -0,0 +1,144 @@
...
+
+	memdb "github.com/hashicorp/go-memdb"
+	"github.com/kong/go-database-reconciler/pkg/utils"
+)
+
+const (
+	licenseTableName = "license"
+)
+
+var licenseTableSchema = &memdb.TableSchema{
```
pmalek, line 15: I wonder what are others' thoughts on this but I'd personally change this to a `func` returning this variable. This way we don't have issues with sharing the same underlying memory area.

pkg/state/license.go
```diff
@@ -0,0 +1,144 @@
...
+type LicensesCollection collection
+
+func getLicense(txn *memdb.Txn, id string) (*License, error) {
+	res, err := multiIndexLookupUsingTxn(txn, licenseTableName, []string{"id"}, id)
+	if err != nil {
+		return nil, err
+	}
+	l, ok := res.(*License)
+	if !ok {
+		panic(unexpectedType)
```
pmalek, line 36: Why not return an error?

tests/integration/sync_test.go
```diff
@@ -5183,3 +5185,124 @@ func Test_Sync_ConsumerGroupConsumerFromUpstream(t *testing.T) {
...
+			CurrentState: currentState,
+			TargetState:  stateWithoutLicense,
+
+			KongClient:      client,
+			IncludeLicenses: false,
+		})
+		require.NoError(t, err)
+
+		stats, errs, changes := syncer.Solve(ctx, 1, false, true)
+		require.Len(t, errs, 0, "Should have no errors in syncing")
```
pmalek, line 5271: FYI there's also `Empty()` if you'd like to use it: https://pkg.go.dev/github.com/stretchr/testify/require#Empty

tests/integration/sync_test.go
```diff
@@ -5183,3 +5185,124 @@ func Test_Sync_ConsumerGroupConsumerFromUpstream(t *testing.T) {
...
+		logEntityChanges(t, stats, changes)
+
+		newState, err := fetchCurrentState(ctx, client, deckDump.Config{IncludeLicenses: true})
+		require.NoError(t, err)
+		licenses, err := newState.Licenses.GetAll()
+		require.NoError(t, err)
+		require.Equal(t, 1, len(licenses))
```
pmalek, line 5278:
```suggestion
		require.Len(t, licensed, 1)
```

tests/integration/sync_test.go
```diff
@@ -5183,3 +5185,124 @@ func Test_Sync_ConsumerGroupConsumerFromUpstream(t *testing.T) {
...
+		require.NoError(t, err)
+		stats, errs, changes := syncer.Solve(ctx, 1, false, true)
+		require.Len(t, errs, 0, "Should have no errors in syncing")
+		logEntityChanges(t, stats, changes)
+
+		newState, err := fetchCurrentState(ctx, client, deckDump.Config{IncludeLicenses: true})
+		require.NoError(t, err)
+		licenses, err := newState.Licenses.GetAll()
+		require.NoError(t, err)
+		require.Equal(t, 0, len(licenses))
```
pmalek, line 5306:
```suggestion
		require.Empty(t, licenses)
```

pmalek (approved):

pkg/state/license.go
randmonkey: This is kept the same as other types.

tests/integration/sync_test.go
randmonkey: Done

pkg/file/builder.go
```diff
@@ -967,6 +973,29 @@ func (b *stateBuilder) vaults() {
...
 
+func (b *stateBuilder) licenses() {
+	if b.err != nil {
+		return
+	}
+
+	for _, l := range b.targetContent.Licenses {
+		l := l
+		if utils.Empty(l.ID) {
+			license, err := b.currentState.Licenses.Get(*l.ID)
```
pmalek, line 984: If `l.ID` is empty we use it in `Get()`? I'm not sure I understand the purpose here 🤔 Might be worth adding a comment about this and/or changing this.

randmonkey: Fixed. This was copied from other types but `license` only have one identifier `ID`, so I think the proper way to process when ID is not given is to generate an ID.

pmalek (approved):

Issue #22, opened 2023-11-30
Refactor console output


rainest:
The original deck design of these libraries had very limited logging. It did not use a logging library, and instead uses a [basic wrapper around `fmt.Println()`](https://github.com/Kong/go-database-reconciler/blob/v1.1.0/pkg/cprint/color.go) to output directly to stdout/stderr. It has a basic [global on/off toggle](https://github.com/Kong/go-database-reconciler/blob/v1.1.0/pkg/cprint/color.go#L13) at the library level, though some consumers further add their own wrappers to simulate additional log levels.

To better make these packages a library, we should instead use a logging library or return rich structures to allow consumers to choose their desired log level and behavior. Library features that currently use `cprint` should instead accept a logger from their instantiator and send log lines to it an appropriate verbosity level, or should eschew logging altogether in favor of returning data that callers can handle as they wish.

There are currently 3 uses of `cprint` in this library:

- The `diff` packages prints updated resource lines and possibly JSON diffs.
- The `types` basic-auth handler has a warning it prints when you attempt to use basic-auths with deck.
- The `util` package can print a warning called from `file.Builder.build()` if routes use regex syntax incompatible with their version.

The nature of these uses push me towards saying using returned values rather than accepting a logger. The full JSON diffs offered by `diff` do not readily work with most logging systems, since they're ideally printed over multiple lines for human consumption. These probably require handling (outside the normal logging system) downstream to present effectively. Continuing to print these to stderr makes sense for the deck CLI, but not so much for KIC--KIC does currently dump these to log output at higher verbosity, but it'd probably make more sense to place these in a separate file accessible via the config diagnostic API.

The warning print should probably be handled via initial scans of the input YAML. The deck CLI can scan input to see if basic-auths are present at all and print that warning (though weirdly as I read it we currently only print these on delete actions? idk why--this feature maybe needs further review), and can scan for regexes that don't match the Kong version similarly. KIC simply does not generate the wrong regex type, and has no issues using basic-auth that I know of.

Acceptance:

- [ ] go-database-reconciler no longer provides the `cprint` package. Any downstream consumers (likely only deck) provide their own instead if its functionality is desired.
- [ ] Functionality that currently uses cprint instead returns rich values that it would have previously formatted and output itself.

Tieske:
Looking at the landscape we have now;

- database reconciler
- go-apiops
- kic (depending on database reconciler)
- deck (depending on database reconciler, and go-apiops)

[...]

PR #76, opened 2024-03-30
Add Syncer result chan

rainest:
# What this does/why we need it

Adds a new result channel to `diff.Syncer`. When enabled, the Syncer disables its traditional console output and instead sends structs describing its actions and errors encountered per entity to an exposed channel.

Downstream library consumers can use this to associate actions and errors with entities they manage. KIC can, for example, use these to generate error Events for the Kubernetes resources associated with an entity or update Programmed status on a per-resource basis (currently it only updates resource status based on the latest complete configuration applied, which may not accurately reflect Kong state during partial applies).

Existing consumers (namely deck) will not see any change in behavior. When disabled, traditional streaming event logs print to the console as usual.

# Special notes

Derived from initial work in https://github.com/Kong/go-database-reconciler/pull/30. The attempt there raised concurrency concerns and difficulties extracting cprint from the library due to test dependencies.

Draft PoC of usage in https://github.com/Kong/kubernetes-ingress-controller/pull/5785


pkg/diff/diff.go
```diff
@@ -90,6 +145,9 @@ type Syncer struct {
 	includeLicenses bool
 
 	isKonnect bool
+
+	// enableEntityActions enables entity actions and disables direct output prints.
+	enableEntityActions bool
```
randmonkey, line 150: How about renaming it and adding more comments to tell it enables a **CHANNEL** that requires the caller to consume it?

pkg/diff/diff.go
pmalek, line 1: `Solve()` deserves a bit of a refactor to at the very least move the separate CRUD related actions to their own respective functions and perhaps carving out function definitions like `recordOp etc`

rainest: Not much disagreement from me there; a lot of the internals of this are difficult to understand and debug, but have remained that way for lack of impetus to change what works. There is an impetus to change some parts here because we can't parse errors into events otherwise (and would like better control over logging downstream).

Is there any particular additional refactor that's relevant to this change? If not I'd say record a refactor issue and handle it separately to limit the scope of this.

pkg/diff/diff.go
rainest: This is set through the options, where the comment mentions as much, but I've updated it so both of them do.

pkg/diff/diff.go
```diff
@@ -526,32 +591,61 @@ func (sc *Syncer) Solve(ctx context.Context, parallelism int, dry bool, isJSONOu
...
-			if err != nil {
-				return nil, err
-			}
-			if isJSONOut {
-				output.Updating = append(output.Updating, item)
+			if sc.enableEntityActions {
+				actionResult.Action = UpdateAction
+				if err != nil {
+					actionResult.Error = err
+					sc.ResultChan <- actionResult
```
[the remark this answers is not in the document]
rainest: That is some hideous syntax :neutral_face:  92862777a2623e7ec564772aa1ab29c103b8b119

rainest:
The [latest gateway dev failure](https://github.com/Kong/go-database-reconciler/files/14845467/fail.txt) is due to a [new key-auth field](https://github.com/Kong/kong-ee/blame/75828130e643e3b9e34be12de7c55a7d317a6996/kong/plugins/key-auth/schema.lua#L31).

I'm trying to think of how I want to handle that. The pseudo-golden tests are quite clunky re version differences, and I don't exactly want to make yet another `Test_Sync_ConsumerGroupsScopedPlugins_After370` after `Test_Sync_ConsumerGroupsScopedPlugins_After350` that's basically identical aside from the one field :neutral_face:

pkg/diff/diff.go
```diff
@@ -338,13 +404,12 @@ func (sc *Syncer) wait() {
 }
 
 // Run starts a diff and invokes d for every diff.
```
pmalek, line 406:
```suggestion
// Run starts a diff and invokes action for every diff.
```

pkg/diff/diff.go
```diff
@@ -68,9 +122,10 @@ type Syncer struct {
 	processor     crud.Registry
 	postProcessor crud.Registry
 
-	eventChan chan crud.Event
-	errChan   chan error
-	stopChan  chan struct{}
+	eventChan  chan crud.Event
+	errChan    chan error
+	stopChan   chan struct{}
+	ResultChan chan EntityAction
```
pmalek, line 128: I just noticed that this field is the only one that would become exported.

While this would work I'd say that it'd be better to return this channel via a method. Even better (but probably out of scope for this PR) introduce a subscription mechanism where each interested party would get its own notification channel.

For now I believe just returning a channel via a method would do so that users do not need to mess with `Syncer`'s fields.

pkg/diff/diff.go
pmalek: Nothing in particular but I thought it's worth pointing out that `Solve` became a bit hard to grasp.

Do you mind creating an issue for that?

pmalek: - https://github.com/Kong/go-database-reconciler/blob/95709ed946b8a327ddd3fa76908df6a6dac0dc32/pkg/diff/diff.go#L562-L582 can be put aside to not clutter up the `func`
- the CRUD switch could become 2 branches, each handled by a dedicated func. Perhaps that could even be extended to the actual `Do` section as well
  - 1 for `sc.enableEntityActions`
  - 1 for `else`

At the very least those 2

pkg/diff/diff.go
rainest: Method's fine, though I doubt we're likely to refactor around multiple Syncer subscribers anytime soon. We don't even persist the things across update operations, and the design til now has essentially been "library clients create and are the exclusive owners of their Syncers". Offhand I don't know of any use cases where we'd want to refactor it for multiple clients and/or persistence, but that'd be the driving factor for such a change should it arise.

pkg/diff/diff.go
rainest: The `else` (original behavior) should go away eventually. The plan is to add this first, then convert deck to it and extract the stats/console output functionality to the deck client. The test structure for deck+GDR still results in some chicken-egg order problems if you try to rip everything out at once, so I've added TODOs attached to #22. We may defer some of those into separate issues, but at present the hope is to clear them out as part of the broader console output refactor.

rainest:
The pseudo-golden tests are both unable to deal with new entity fields for new versions and are structured so that any version-specific carveouts are basically impossible AFAICT :frowning_face: 

We do not have version gates for parts of test functionality, only the blanket `runWhen()` skips. Even if we did, we'd still have to maintain separate expected states or inject the extra field into one, which is still cumbersome.

We could alternately delete fields out of the actual state in `testKongState()`, but this would be similarly cumbersome.

We can sort of ignore fields in general, but again, same version gate caveats, and you furthermore need to use an incredibly blunt instrument due to `kong.Configuration` being an arbitrary `map[string]interface{}`: [hack.diff.txt](https://github.com/Kong/go-database-reconciler/files/14859611/hack.diff.txt) works, but yeesh.

So, sadly, went ahead and just made a 3.7 variant of the tests :neutral_face:

pmalek:
The CI seems to be broken for quite some time. Not sure how this went unnoticed. Anyway #79 should fix it.

We can then pull in that fix to this PR to properly run tests.

pkg/diff/diff.go
```diff
@@ -45,6 +56,49 @@ type EntityChanges struct {
...
+const (
+	// CreateAction is the ReconcileAction used when a target object did not exist in the current state and was created.
+	CreateAction = ReconcileAction("create")
+	// UpdateAction is the ReconcileAction used when a target object exists in the current state and was updated.
+	UpdateAction = ReconcileAction("update")
+	// DeleteAction is the ReconcileAction used when a current object exists in the target state and was deleted.
+	DeleteAction = ReconcileAction("delete")
+
+	// eventBuffer is the number of events to buffer in the various syncer channels.
+	eventBuffer = 10
```
randmonkey, line 75: Is there any reason for setting `eventBuffer` to be this constant?

randmonkey (approved):

pkg/diff/diff.go
rainest: Not really. It was an existing value from https://github.com/Kong/go-database-reconciler/pull/76/files#diff-00282c65a618a9ea64cdb99da5137663dc5773f2c3fd8c37ed2e9a99f3d67f09L347; I just moved the definition.

I expect that it's relatively small to avoid overloading the admin API with a ton of requests in flight if the DB is slow to complete them. Could make sense to make it configurable (a library const isn't great for our new client/library model), but we haven't found that we need to configure it so far.

PR #82, opened 2024-04-08
ci: fix `make clean` and refactor

flrgh:
The integration cleanup script (`.ci/clean_kong.sh`) had a couple warts:

* the `--filter Name=<name>` arg performs substring matching, so if you have an unrelated container named `kong-foo` running, it would enter the wrong branch and throw an error
* manually tracking docker resources for cleanup gets pretty tedious

I updated it to filter resources based on a known docker label instead.

This involved touching a few different files to add docker labels and such, so I wound up DRY-ing up some of the CI setup scripts along the way.

### DIFF
--- a/.ci/clean_kong.sh
+++ b/.ci/clean_kong.sh
@@ -1,29 +1,23 @@
 #!/bin/bash
 
-set -e
+set -euo pipefail
 
-NETWORK_NAME=deck-test
+source ./.ci/lib.sh
 
-PG_CONTAINER_NAME=pg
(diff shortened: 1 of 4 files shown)


pmalek (approved):
Looks good. Left some minor nits but feel free to disregard those.

> This makes local integration testing better

Might be better to include the specific issue that you're trying to solve with this PR.

flrgh:
@pmalek I think applying your suggestions and creating another commit also invalidated your approval :upside_down_face:. Can you re-approve please?

pmalek (approved):

PR #93, opened 2024-05-22
fix: correct plugins config comparison

GGabriele:
This commit fixes a bug with the way Plugins config are compared: when unsorted nested arrays are present, `reflect.DeepEqual` considers the equal unsorted arrays as unequal, triggering the plugin update and showing misleading diffs.

This commit is introducing a few helper functions to recursively sort arrays within the Kong Plugins configs.

### Summary

SUMMARY_GOES_HERE

### Full changelog

* [Implement ...]
* [Fix ...]

### Issues resolved

Fix #XXX

### Documentation

### Testing

- [ ] Unit tests
- [ ] E2E tests


pkg/state/types.go
```diff
@@ -552,6 +555,49 @@ func (p1 *Plugin) EqualWithOpts(p2 *Plugin, ignoreID,
...
+	sort.Strings(strSlice)
+
+	// Convert back to []interface{}
+	sortedSlice := make([]interface{}, len(strSlice))
+	for i, v := range strSlice {
+		sortedSlice[i] = v
+	}
+
+	return sortedSlice
+}
```
pmalek, line 575: How about using a type that can be sorted and thus not allocating unnecessarily?

Something like this: d7c9b78036793944648cb0a842000018d0635975

WDYT?

pmalek: Pushed it directly to this PR: https://github.com/Kong/go-database-reconciler/pull/93/commits/d7c9b78036793944648cb0a842000018d0635975.

PTAL

Tieske (approved):

PR #101, opened 2024-06-06
fix: detect filter chain support from API response

flrgh:
This updates the business logic that is used to decide if (wasm) filter chains are supported by Kong. Instead of relying on the caller to set `dump.Config.IsFilterChainsSupported`, we check the error response code from `GET /filter-chains` and decide how to proceed from there.

---

See also: https://github.com/Kong/deck/pull/987 specifically: https://github.com/Kong/deck/pull/987#discussion_r1629376245

KAG-4005

### DIFF
--- a/pkg/dump/dump.go
+++ b/pkg/dump/dump.go
@@ -42,9 +42,6 @@ type Config struct {
 
 	// IsConsumerGroupScopedPluginSupported
 	IsConsumerGroupScopedPluginSupported bool
-
-	// IsFilterChainsSupported
-	IsFilterChainsSupported bool
 }
 
 func deduplicate(stringSlice []string) []string {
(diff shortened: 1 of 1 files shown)


GGabriele (approved):

pkg/dump/dump.go
```diff
@@ -255,18 +252,25 @@ func getProxyConfiguration(ctx context.Context, group *errgroup.Group,
...
-			filterChains, err := GetAllFilterChains(ctx, client, config.SelectorTags)
-			if err != nil {
-				return fmt.Errorf("filter chains: %w", err)
-			}
-			state.FilterChains = filterChains
-			return nil
-		})
-	} else {
+	group.Go(func() error {
 		state.FilterChains = make([]*kong.FilterChain, 0)
```
pmalek, line 256: We don't really need to set this here and allocate the slice, right?

Line 271 will do I guess.

flrgh: You're right--we don't strictly _need_ to allocate here. I thought it better to have the "set a default value" code at the top rather than hiding it 3 levels deep in the "we got an error, but it's okay" branch, but I have no objections to moving it.

PR #109, opened 2024-07-01
fix: correctly compare compressed/expanded ipv6 in targets

GGabriele:
When syncing a target with a compressed IPv6 (for example `[2001:db8:fd73::e]:1326`) decK can successfully create it but it fails to detect its existence when pulling the config from the API because the CP expands the address (for example to `[2001:0db8:fd73:0000:0000:0000:0000:000e]:1326`) causing decK to attempt to recreate the target and failing because the target already exists.

This commit fixes this by using some helpers to expand the ipv6 before trying to sync/dump.

### Summary

SUMMARY_GOES_HERE

### Full changelog

* [Implement ...]
* [Fix ...]

### Issues resolved

Fix #XXX

### Documentation

### Testing

- [ ] Unit tests
- [ ] E2E tests

### DIFF
--- a/pkg/file/builder.go
+++ b/pkg/file/builder.go
@@ -2,10 +2,10 @@ package file
 
 import (
 	"context"
-	"encoding/hex"
 	"errors"
 	"fmt"
 	"net"
+	"net/netip"
 	"reflect"
 	"regexp"
 	"sort"
@@ -1052,20 +1052,17 @@ func isIPv6(hostname string) bool {
 //
 // from ::1 to 0000:0000:0000:0000:0000:0000:0000:0001.
 func expandIPv6(address string) string {
-	ip := net.ParseIP(address).To16()
-	dst := make([]byte, hex.EncodedLen(len(ip)))
-	hex.Encode(dst, ip)
-	var final string
-	for i := 0; i < len(dst); i += 4 {
-		final += fmt.Sprintf("%s:", dst[i:i+4])
-	}
-	if len(final) == 0 {
+	addr, err := netip.ParseAddr(address)
+	if err != nil {
 		return ""
 	}
-	// remove last colon
-	return final[:len(final)-1]
+	addr.StringExpanded()
+	return addr.StringExpanded()
 }
 
+// normalizeIPv6 normalizes an ipv6 address to the format [address]:port.
+// for example:
+// from ::1 to [0000:0000:0000:0000:0000:0000:0000:0001]:8000.
 func normalizeIPv6(target string) (string, error) {
 	ip := target
 	port := "8000"
@@ -1138,6 +1135,15 @@ func (b *stateBuilder) upstreams() {
 func (b *stateBuilder) ingestTargets(targets []kong.Target) error {
 	for _, t := range targets {
 		t := t
+
+		if t.Target != nil && isIPv6(*t.Target) {
+			normalizedTarget, err := normalizeIPv6(*t.Target)
+			if err != nil {
+				return err
+			}
+			t.Target = kong.String(normalizedTarget)
+		}
+
 		if utils.Empty(t.ID) {
 			target, err := b.currentState.Targets.Get(*t.Upstream.ID, *t.Target)
 			if errors.Is(err, state.ErrNotFound) {
@@ -1150,13 +1156,6 @@ func (b *stateBuilder) ingestTargets(targets []kong.Target) error {
 		}
 		utils.MustMergeTags(&t, b.selectTags)
 		b.defaulter.MustSet(&t)
-		if t.Target != nil && isIPv6(*t.Target) {
-			normalizedTarget, err := normalizeIPv6(*t.Target)
-			if err != nil {
-				return err
-			}
-			t.Target = kong.String(normalizedTarget)
-		}
 		b.rawState.Targets = append(b.rawState.Targets, &t)
 	}
 	return nil
--- a/pkg/file/builder.go
+++ b/pkg/file/builder.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"net"
 	"net/netip"
 	"reflect"
 	"regexp"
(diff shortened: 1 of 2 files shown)


pkg/file/builder.go
```diff
@@ -1033,6 +1037,65 @@ func (b *stateBuilder) rbacRoles() {
...
 }
 
+var (
+	IPv6HasPortPattern    = regexp.MustCompile(`\]\:\d+$`)
+	IPv6HasBracketPattern = regexp.MustCompile(`\[\S+\]$`)
+)
+
+func isIPv6(hostname string) bool {
+	parts := strings.Split(hostname, ":")
+	return len(parts) > 2
```
pmalek, line 1047: This can be prone to errors or to be misused. Why not use https://pkg.go.dev/net/netip#ParseAddr and https://pkg.go.dev/net/netip#Addr.Is6 ?

pkg/file/builder.go
```diff
@@ -1033,6 +1037,65 @@ func (b *stateBuilder) rbacRoles() {
...
+}
+
+// expandIPv6 decompress an ipv6 address into its 'long' format.
+// for example:
+//
+// from ::1 to 0000:0000:0000:0000:0000:0000:0000:0001.
+func expandIPv6(address string) string {
```
pmalek, line 1054: Isn't this basically what https://pkg.go.dev/net/netip#Addr.StringExpanded does?

pkg/file/builder.go
```diff
@@ -1033,6 +1037,65 @@ func (b *stateBuilder) rbacRoles() {
...
+		final += fmt.Sprintf("%s:", dst[i:i+4])
+	}
+	if len(final) == 0 {
+		return ""
+	}
+	// remove last colon
+	return final[:len(final)-1]
+}
+
+func normalizeIPv6(target string) (string, error) {
```
pmalek, line 1069: This one doesn't have a godoc with an example of a UT so I'm not sure what it does but perhaps it can also be replaced with functions from `netip`?

pkg/file/builder.go
GGabriele: As silly as it may look, this is enough to catch all combinations of targets in IPv6 format, including the one with port and brackets like `[2600:1f16:1784:fa03:fd73::e]:1326`, which the library wouldn't consider as v6

pkg/file/builder.go
GGabriele: Yes! I didn't know about `net/netip`, thanks!

pkg/file/builder.go
GGabriele: Added some description, let me know if that helps.

pkg/file/builder.go
```diff
@@ -1033,6 +1037,62 @@ func (b *stateBuilder) rbacRoles() {
...
+// for example:
+//
+// from ::1 to 0000:0000:0000:0000:0000:0000:0000:0001.
+func expandIPv6(address string) string {
+	addr, err := netip.ParseAddr(address)
+	if err != nil {
+		return ""
+	}
+	addr.StringExpanded()
+	return addr.StringExpanded()
```
pmalek, line 1060:
```suggestion
	return addr.StringExpanded()
```

pkg/file/builder.go
```diff
@@ -1033,6 +1037,62 @@ func (b *stateBuilder) rbacRoles() {
...
+}
+
+// normalizeIPv6 normalizes an ipv6 address to the format [address]:port.
+// for example:
+// from ::1 to [0000:0000:0000:0000:0000:0000:0000:0001]:8000.
+func normalizeIPv6(target string) (string, error) {
```
pmalek, line 1066: Couldn't we also use `StringExpanded()` here? Possibly with a port?

pkg/file/builder.go
pmalek: Wouldn't https://pkg.go.dev/net/netip#ParseAddrPort work for this then?

I'm just worried that this func is very easy to abuse.

`isIPv6("this:is:nuts:!")` will also return `true`. It will also return true for `1:2:3:4` which http://sqa.fyicenter.com/1000334_IPv6_Address_Validator.html#Result reports as invalid.

pkg/file/builder.go
GGabriele: ops :D

pkg/file/builder.go
GGabriele: `ParseAddrPort` works if the port is always there, but it would fail to parse things like `::1` (would work with `[::1]:8000` though).

If you really feel strong about this we can improve this validation, but generally decK (or this library) doesn't do much input validation other than what's part of the JSON schemas from go-kong. 

For the records though, those bogus entries would still fail validation at a later stage:

```
$ ./deck gateway sync kong.yaml
Error: building state: invalid ipv6 address 1:2:3:4

$ ./deck gateway sync kong.yaml
Error: building state: invalid ipv6 address this:is:nuts:!
```

pmalek: Understood that this is validated on other levels.

I'm just suggesting to consider improving the reliability of this while we're at it

For instance something like this should work just fine and will work for addresses with and without ports

```
func IsIPv6(str string) bool {
	addr, err := netip.ParseAddrPort(str)
	if err != nil {
		addr, err := netip.ParseAddr(str)
		fmt.Printf("error parsing address: %v\n", err)
		if err != nil {
			fmt.Printf("error parsing address port: %v\n", err)
			return false
		}
		return addr.Is6()
	}

	return addr.Addr().Is6()
}
```

pmalek (dismissed):
Code looks good and serves it's purpose 👍 

Feel free to address the comments or merge as is if time is of the essence.

pkg/file/builder.go
GGabriele: This still would fail to parse something like `[::1]`, as well as detecting bogus patterns like `[invalid:ipv6::address]:1326`: the function would rightly return `false`, but this is not just `false`, it's invalid. At this point we would need add extra handling and validation for these cases too. I think this is going to become more complex and fragile overall.

I'm renaming this function to `hasIPv6Format` which better represent its purpose: https://github.com/Kong/go-database-reconciler/pull/109/commits/6764fd2987b96ab1ef282abdd31c070dc3fa8b83

pmalek (approved):

PR #117, opened 2024-07-22
test: Valid expression routes building

Prashansa-K:
Test for PR: https://github.com/Kong/go-database-reconciler/pull/116

One test for ingest routes are separated into two to deal with different kinds of routes - expression and traditional.
Added assertion to ensure that `regex_priority` is not present in expression route but present in the traditional routes.

### DIFF
--- a/pkg/file/builder_test.go
+++ b/pkg/file/builder_test.go
@@ -3064,7 +3064,6 @@ func Test_stateBuilder_ingestRouteKonnectTraditionalRoute(t *testing.T) {
 			wantState: &utils.KongRawState{
 				Routes: []*kong.Route{
 					{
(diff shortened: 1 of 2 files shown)


pkg/file/builder_test.go
```diff
@@ -3074,6 +3074,43 @@ func Test_stateBuilder_ingestRouteKonnect(t *testing.T) {
 				},
 			},
 		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			ctx := context.Background()
+			b := &stateBuilder{
+				currentState: tt.fields.currentState,
+				isKonnect:    true,
```
GGabriele, line 3083: Can we also test with `isKonnect: false`?

GGabriele (dismissed):

GGabriele (approved):

PR #134, opened 2024-08-26
fix: plugin shown as global despite a consumer-group present

Prashansa-K:
Prior to this fix, deck was reporting the plugins as global even if a consumer-group was attached to them. This was only a output problem, not a association problem. The fix checks if a consumer-group exists or not before reporting any plugin as global.

Fixes: https://github.com/Kong/deck/issues/1005

### Documentation

### Testing

- [x] Unit tests
- [ ] E2E tests

### DIFF
--- a/pkg/state/types_test.go
+++ b/pkg/state/types_test.go
@@ -663,3 +663,49 @@ func TestDeepEqualWithSorting(t *testing.T) {
 		t.Errorf("expected maps to be equal, but they are not")
 	}
 }
+
+func TestConsole(t *testing.T) {
+	assert := assert.New(t)
+
+	var p1 Plugin
+	p1.ID = kong.String("foo")
+	p1.Name = kong.String("foo-plugin")
+
+	// Testing for global
+	expected := "foo-plugin (global)"
+	actual := p1.Console()
+	assert.Equal(expected, actual)
+
+	// Adding a service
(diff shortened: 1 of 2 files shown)


pkg/state/types_test.go
```diff
@@ -663,3 +663,49 @@ func TestDeepEqualWithSorting(t *testing.T) {
 		t.Errorf("expected maps to be equal, but they are not")
 	}
 }
+
+func TestConsole(t *testing.T) {
```
czeslavo, line 667: To make sure it's clear what the test covers.
```suggestion
func TestPluginConsole(t *testing.T) {
```

pkg/state/types_test.go
```diff
@@ -663,3 +663,49 @@ func TestDeepEqualWithSorting(t *testing.T) {
...
+	assert.Equal(expected, actual)
+
+	// Making everything nil
+	p1.Service = nil
+	p1.Route = nil
+	p1.Consumer = nil
+	p1.ConsumerGroup = nil
+	expected = "foo-plugin (global)"
+	actual = p1.Console()
+	assert.Equal(expected, actual)
```
czeslavo, line 710: Please consider converting this into a table-driven test with a separate test case for every commented section.

czeslavo (commented):
👍 Just two tiny nits in tests.

czeslavo (approved):

PR #139, opened 2024-09-03
fix(sync): bump go-kong + fill auto fields with nil

samugi:
### Summary

bump go-kong to 0.59.0

use go-kong utils to fill auto fields with nil in the config sent to the CP.
The rest of the default values are still not filled in the configuration that is sent out, as they are meant to be filled by the CP itself.

### Full changelog

* [Implement ...]
* [Fix ...]

### Issues resolved

https://konghq.atlassian.net/browse/KAG-5317

### Documentation

### Testing

- [ ] Unit tests
- [ ] E2E tests


pkg/diff/diff.go
```diff
@@ -620,12 +620,19 @@ func (sc *Syncer) Solve(ctx context.Context, parallelism int, dry bool, isJSONOu
...
+					return nil, fmt.Errorf("failed processing auto fields: %w", err)
+				}
+
+				// only fill auto fields for the configuration sent to Kong
+				// this is done because we want to avoid Kong to auto generate fields, which
+				// would make decK's configuration no longer fully "declarative"
+				if err := kong.FillPluginsDefaultsWithOpts(&plugin.Plugin, schema, kong.FillRecordOptions{
+					FillDefaults: false,
+					FillAuto:     true,
+				}); err != nil {
```
bungle, line 635: I really don't still understand what this has to do with being `fully declarative`. Why can we skip filling `defaults`  but not skipping filling `auto` fields? Is it because `defaults` are `static` in schema (but can change between Kong releases) and that `auto` fields are usually functions that return e.g. current timestamp or random string?

If you send auto field that is filled with `null` (Go nil) to Kong and that field is `required=true`, then the request will fail. Also if you send auto field that is filled with  `null` (Go nil) that is `required=false`, then Kong will not fill in the default.

This can lead to many potential, even security issues.

Let's look for one example:
https://github.com/Kong/kong-ee/blob/master/plugins-ee/openid-connect/kong/plugins/openid-connect/schema.lua#L1973-L1978

```lua
{
  cache_tokens_salt = {
    description = "Salt used for generating the cache key that is used for caching the token endpoint requests.",
    required = false,
    type = "string",
    auto = true,
  },
},
```         

And user YAML looks like:
```yaml
plugins:
  - name: openid-connect
     issuer: http://...
```

The user didn't enter `cache_tokens_salt` which to Kong means that Kong needs to generate it. So by default Kong does generate different `cache_tokens_salt` on each plugin (this is potential security issues if it doesn't do it). Now if people want to share the salt between plugin instances, they can either set it forcibly to `null` in their YAML or same value like `"my-salt"`. Then the cache will be shared between two plugin instances. But this is not default behavior. What decK does here is that it forcibly adds `cache_tokens_salt=null` before it sends configuration to Kong, which means that every OpenID Connect plugin that is in decK will share the same cache, which again is a potential security issue.

I do not see point in this `"let's not fill defaults, but let's fill autos"` so that `decK` can be fully declarative. When it comes to being fully declarative I only think the user source YAML. I do understand that on each sync the Kong may generate different `cache_tokens_salt`, but that is the right behavior as decK uses `PUT`.

bungle (changes requested):
I still do not think this is good to have.

pkg/diff/diff.go
mheap: I actually agree with you about `null` for things like OIDC. However, this is how `deck` works today and in the interest of shipping the "don't fill defaults" behaviour I think this is a good tradeoff to make.

Let's continue to talk about explicit `null` for `auto` fields, but this PR allows us to continue shipping whilst minimising risk of breaking changes.

bungle: Now let's look another:
https://github.com/Kong/kong/blob/master/kong/db/schema/entities/plugins.lua#L21C29-L21C45
```lua
{ updated_at = typedefs.auto_timestamp_s },
```

where `auto_timestamp_s` is defined as:
```lua
typedefs.auto_timestamp_s = Schema.define {
  type = "integer",
  timestamp = true,
  auto = true,
  description = "An integer representing an automatic Unix timestamp in seconds."
}
```

Again here `required=false`, which means that if decK fill `null` the `null` will be stored in database for `updated_at`. Whereas if you used normal admin api, Kong would fill in the current timestamp to `updated_at` on same _USER_  input (with decK the user written YAML that does not contain `updated_at` and with Kong Admin API a request that doers not contain `updated_at`).  Here decK fills it, and it again diverges from Kong default behavior, and user intent and user YAML.

bungle: @mheap, if ^ doesn't change your decision about this, please mark this as resolved, and I can approve the PR.

samugi: maybe we could/should have a follow-up to differentiate the behavior between required and non-required auto fields and only fill the former. This PR is more similar to the previous behavior so somewhat safer than just going all in and not filling any values.

bungle: @samugi yes, but filling `required=true` auto-fields with `null` (Go nil) will just lead to sync error on validation.

mheap: This is the current behaviour, which I'd like to keep. We should continue the discussion if this behaviour should be changed

randmonkey (approved):

randmonkey (approved):

PR #153, opened 2024-12-09
feat: ignoring konnect-managed entities with deck

Prashansa-K:
### Summary

As of now, deck supports ignoring plugins
that are konnect-managed. We are adding further
support to ignore any entity that may have the
tag "konnect-managed" added to it. Konnect
adds these tags to some entities and they
are treated as immutable. Thus, we wish to
ignore any deck operations on them.

With this change, a deck dump would not
include the "konnect-managed" entities.
deck sync/diff would also ignore these entities.

### Issues resolved

For: https://github.com/Kong/deck/issues/1442

### Documentation

### Testing

- [x] Unit tests
- [ ] E2E tests


mheap:
@Prashansa-K Do you think excluding from `dump` is the right choice here? If we ignore on `diff` and `sync` there may be a use case to return them in `dump` so that users can see the complete picture.

The downside is that if people edit those fields in their config file, they won't be synced back. I can see both sides here.

Prashansa-K:
> @Prashansa-K Do you think excluding from `dump` is the right choice here? If we ignore on `diff` and `sync` there may be a use case to return them in `dump` so that users can see the complete picture.

@mheap Aren't the entities that are tagged as "konnect-managed" internal-use entities? If they are, we should not show them in dump.

Also, how are the other "konnect-managed" entities different from "konnect-managed" plugins? We currently opt not to dump the konnect-managed plugins.

mheap:
Prior art here is a good example. Let's not dump them 👍

GGabriele (approved):
LGTM, thanks!

PR #154, opened 2024-12-19
feat: added support for degraphql_routes

Prashansa-K:
### Summary

Adding support for degraphql_routes so that
they can be managed with decK.

Tests added in separate PR: https://github.com/Kong/go-database-reconciler/pull/155
Working on integration tests in the same PR.

**YAML representation:**
```
plugin_entities:
  - type: degraphql_routes
    fields:
      uri: "/contacts"
      query: "query{ contacts { name } }"
      service: my-test-service
```

The above representation is chosen so that other custom-entities can be added in the same way, without changes to parsing/dumping logic. In the code, `FPluginEntity` is used as a superset type that can hold different custom entities.

### Issues resolved

For https://github.com/Kong/deck/issues/1346

### Documentation

### Testing

- [x] Unit tests
- [ ] E2E tests


pkg/state/builder.go
```diff
@@ -416,3 +449,40 @@ func GetKonnectState(rawKong *utils.KongRawState,
 	}
 	return kongState, nil
 }
+
+func buildDegraphqlRouteFromCustomEntity(kongState *KongState, entity map[string]interface{}) (DegraphqlRoute, error) {
+	var degraphqlRoute DegraphqlRoute
+
+	if entity["id"] != nil {
+		degraphqlRoute.ID = kong.String(entity["id"].(string))
```
GGabriele, line 457: In this function, please handle cases in which the type assertions fail. We never know what users can include in their files and failing with a graceful error is always desirable

pkg/state/degraphql_route.go
```diff
@@ -0,0 +1,113 @@
+package state
+
+import (
+	"fmt"
+
+	memdb "github.com/hashicorp/go-memdb"
+	"github.com/kong/go-database-reconciler/pkg/state/indexers"
+)
+
+// DegraphqlRoutesCollection stores and indexes jwt-auth credentials.
```
GGabriele, line 10:
```suggestion
// DegraphqlRoutesCollection stores and indexes degraphql_routes.
```

pkg/state/plugin_entities.go
```diff
@@ -0,0 +1,169 @@
+package state
+
+import (
+	"errors"
+	"fmt"
+
+	memdb "github.com/hashicorp/go-memdb"
+)
+
+// pluginEntitiesCollection stores and indexes key-auth credentials.
```
GGabriele, line 10:
```suggestion
// pluginEntitiesCollection stores and indexes plugin entities.
```

pkg/state/plugin_entities.go
```diff
@@ -0,0 +1,169 @@
...
+	err = txn.Insert(k.PluginEntityType, e)
+	if err != nil {
+		return err
+	}
+
+	txn.Commit()
+	return nil
+}
+
+func (k *pluginEntitiesCollection) deleteCred(txn *memdb.Txn, nameOrID string) error {
```
GGabriele, line 119:
```suggestion
func (k *pluginEntitiesCollection) deleteEntity(txn *memdb.Txn, nameOrID string) error {
```

pkg/state/types.go
```diff
@@ -1791,3 +1791,50 @@ func (l *License) EqualWithOpts(l2 *License, ignoreID, ignoreTS bool) bool {
...
+	// Type of the plugin entity
+	GetPluginEntityType() string
+}
+
+type DegraphqlRoute struct {
+	kong.DegraphqlRoute `yaml:",inline"`
+	Meta
+}
+
+// Identifier returns the ID of the DegraphqlRoute.
```
GGabriele, line 1807:
```suggestion
// GetPluginEntityID returns the ID of the DegraphqlRoute.
```

pkg/state/types.go
```diff
@@ -1791,3 +1791,50 @@ func (l *License) EqualWithOpts(l2 *License, ignoreID, ignoreTS bool) bool {
...
+func (d *DegraphqlRoute) GetPluginEntityID() string {
+	return *d.ID
+}
+
+// Identifier returns the ID of the DegraphqlRoute.
```
GGabriele, line 1812:
```suggestion
// GetPluginEntityType returns the DegraphqlRoute type.
```

pkg/state/degraphql_route.go
```diff
@@ -0,0 +1,113 @@
...
+		return nil, fmt.Errorf("uri/query cannot be empty string")
+	}
+
+	txn := k.db.Txn(false)
+	defer txn.Abort()
+
+	return getDegraphqlRouteByURIQuery(txn, uri, query)
+}
+
+// Add adds a degraphql route credential to DegraphqlRoutesCollection
```
GGabriele, line 66:
```suggestion
// Add adds a degraphql route to DegraphqlRoutesCollection
```

GGabriele (commented):
From a code perspective, this looks good to me, except for a few nits.

I have a few comments from a product perspective though. I see we are following what the [KIC team is doing](https://github.com/Kong/deck/issues/1346#issuecomment-2443589799), but:
- we are calling this `plugin_entities`, but there is nothing related to plugins inside each entity. I understand the reasons because I know what's behind the scenes, but I think users may be confused by the naming
- still on the naming, we usually identify as an `entity` as a single _row_ of a given type. Now saying `I have 10 plugin entities` sounds a bit ambiguous: do you have 10 `plugins` or do you have 10 `plugin_entities`?
- `plugin_entities` is not DB-less compatible. I know decK is not fully db-less compatible because some differences exist on `consumer_groups`, but these differences have been source of friction for our users. Do we intend to further expand on this friction?

Prashansa-K:
> plugin_entities is not DB-less compatible. I know decK is not fully db-less compatible because some differences exist on consumer_groups, but these differences have been source of friction for our users. Do we intend to further expand on this friction?

So, from `kong config db_export` I get this output for degraphql_routes:
```
degraphql_routes:
- id: 73eb90bd-eda7-4c80-aa28-b0d657ea1aa4
  service: 1300b3a5-73c6-4721-a531-4a04ab4cbf13
  updated_at: 1734694480
  created_at: 1734694480
  query: query{ contacts { name } }
  methods:
  - GET
  uri: /contacts
```

DB-less mode identifies `degraphql_routes` but not `plugin_entities`. 
We add the former in the KongState, hence the output. This may cause discrepancy while working in db-less mode.

cc: @mheap

mheap:
I hadn't considered db-less compatibility. This is a very good callout.

How much work would it be to change the format to match DB-less?

Prashansa-K:
> I hadn't considered db-less compatibility. This is a very good callout.
> 
> How much work would it be to change the format to match DB-less?

@mheap As of now, I had considered a common yaml format for all plugin entities and the code was doing the parsing, processing and everything else accordingly. Only the API calls were made to degraphql_routes.

DB-less identifies "degraphql_routes" as a separate entity. So, I would have to rewrite a bunch of things to match it to DB-less. This would also defeat the purpose of using a common way and yaml format to handle all plugin entities.

mheap:
@hbagdi What are your thoughts here? I'd rather not increase the gap between DB-less and Deck formats, but I think this is an instance where we can provide a nicer UX + minimise maintenance cost

mheap:
Let's proceed with the current design rather than mirroring the DB-less format

GGabriele (approved):

PR #155, opened 2024-12-20
chore: added tests for degraphql_routes entity

Prashansa-K:
### Summary

Tests added for new entity: degraphql_routes

### Testing

- [x] Unit tests
- [x] E2E tests


pkg/file/builder_test.go
```diff
@@ -3853,3 +3870,151 @@ func Test_stateBuilder_expressionRoutes_kong370_withKonnect(t *testing.T) {
...
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			b := &stateBuilder{
+				targetContent: tt.fields.targetContent,
+				currentState:  tt.fields.currentState,
+			}
+			b.build()
```
pmalek, line 4016: Any reason we don't check the error returned from `build()`?

pkg/file/builder_test.go
```diff
@@ -3853,3 +3870,151 @@ func Test_stateBuilder_expressionRoutes_kong370_withKonnect(t *testing.T) {
 		})
 	}
 }
+
+func Test_stateBuilder_ingestPluginEntities(t *testing.T) {
+	rand.Seed(42)
```
pmalek, line 3875: This is deprecated as per: https://pkg.go.dev/math/rand#Seed

> Deprecated: As of Go 1.20 there is no reason to call Seed with a random value. Programs that call Seed with a known value to get a specific sequence of results should use New(NewSource(seed)) to obtain a local random generator.

pkg/state/degraphql_route_test.go
```diff
@@ -0,0 +1,278 @@
...
+	degraphqlRoute.ID = kong.String("example")
+	degraphqlRoute.URI = kong.String("/foo")
+	degraphqlRoute.Query = kong.String("query { hello }")
+	degraphqlRoute.Service = &kong.Service{
+		ID: kong.String("some-service"),
+	}
+	degraphqlRoute.Methods = kong.StringSlice("GET")
+
+	err := collection.Add(degraphqlRoute)
+	assert.Nil(err)
```
pmalek, line 150: Nit: I'd consider using `require` for these errors, WDYT?

pkg/state/degraphql_route_test.go
```diff
@@ -0,0 +1,278 @@
...
+
+	populateDegraphqlRoutes(assert, collection)
+
+	degraphqlRoutes, err := collection.GetAll()
+	assert.Nil(err)
+	assert.Equal(5, len(degraphqlRoutes))
+	assert.IsType([]*DegraphqlRoute{}, degraphqlRoutes)
+}
+
+func populateDegraphqlRoutes(assert *assert.Assertions,
```
pmalek, line 213: Nit: this is fine but I'd consider passing the `t *testing.T` instead of the `assert` object as using the latter can potentially (not necessarily) result in an unrelated `assert` being used.

IMHO the former is also more idiomatic. WDYT?

pmalek (commented):
Minor nits, otherwise this looks good.

pkg/file/builder_test.go
Prashansa-K: I have added changes based on the recommendations in rand package. We need that for deterministically generating uuids for the entities.

pkg/file/builder_test.go
```diff
@@ -389,10 +390,29 @@ func existingFilterChainState() *state.KongState {
 	return s
 }
 
+func existingDegraphqlRouteState() *state.KongState {
+	s, _ := state.NewKongState()
```
pmalek, line 394: Why not check the error? I'd suggest we consider adding a `*testing.T` param and `require.NoError(t, err)` here (as well as `t.Helper()` to make sure the correct call site of `existingDegraphqlRouteState` is printed in the logs). WDYT?

pkg/file/builder_test.go
```diff
@@ -3853,3 +3873,240 @@ func Test_stateBuilder_expressionRoutes_kong370_withKonnect(t *testing.T) {
...
+				currentState:  tt.fields.currentState,
+			}
+			_, _, err := b.build()
+			if tt.wantErr {
+				require.NotNil(t, err, "build error is nil")
+				assert.Equal(t, tt.want, b.rawState)
+				return
+			}
+
+			require.Nil(t, err, "build error is not nil")
```
pmalek, line 4108:
```suggestion
			require.NoError(t, err, "build error is not nil")
```

pkg/file/builder_test.go
```diff
@@ -3853,3 +3873,240 @@ func Test_stateBuilder_expressionRoutes_kong370_withKonnect(t *testing.T) {
...
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			b := &stateBuilder{
+				targetContent: tt.fields.targetContent,
+				currentState:  tt.fields.currentState,
+			}
+			_, _, err := b.build()
+			if tt.wantErr {
+				require.NotNil(t, err, "build error is nil")
```
pmalek, line 4103:
```suggestion
				require.Error(t, err, "build error is nil")
```

pkg/state/degraphql_route_test.go
```diff
@@ -0,0 +1,277 @@
...
+	}
+	degraphqlRoute.Methods = kong.StringSlice("GET")
+
+	err := collection.Add(degraphqlRoute)
+	require.Nil(t, err, "error adding degraphql route")
+
+	// Fetch the currently added entity
+	res, err := collection.Get("example")
+	require.Nil(t, err, "error getting degraphql route")
+	assert.NotNil(res)
```
pmalek, line 154: This requires adding `require` instantiation above

```suggestion
	require.NotNil(res)
```

pkg/state/degraphql_route_test.go
```diff
@@ -0,0 +1,277 @@
...
+	collection := degraphqlRoutesCollection()
+
+	var degraphqlRoute DegraphqlRoute
+	degraphqlRoute.ID = kong.String("example")
+	degraphqlRoute.URI = kong.String("/foo")
+	degraphqlRoute.Query = kong.String("query { hello }")
+	degraphqlRoute.Service = &kong.Service{
+		ID: kong.String("some-service"),
+	}
+	degraphqlRoute.Methods = kong.StringSlice("GET")
```
pmalek, line 146: Any reason not to instantiate it like so:

```suggestion
	var degraphqlRoute := DegraphqlRoute{
		...
	}
```

?

pmalek (commented):
Minor nits here and there for consideration

PR #157, opened 2025-01-06
fix: returning a validation error in case of wrong configuration.

Prashansa-K:
As of now, while using deck to create a
route-scoped plugin with a `service` field
defined, the sync goes through the first
time but errors out subsequently. Under a
route, we do not want the user to add the
`service` field for a plugin. Thus, we are
checking for this specific scenario and
erroring out, instead of letting the sync
action go through.

For: https://github.com/Kong/deck/issues/1353

### Documentation

### Testing

- [ ] Unit tests
- [ ] E2E tests

### DIFF
--- a/pkg/file/builder.go
+++ b/pkg/file/builder.go
@@ -19,6 +19,13 @@ import (
 
 const ratelimitingAdvancedPluginName = "rate-limiting-advanced"
 
+const (
+	primaryRelationConsumer      = "consumer"
+	primaryRelationConsumerGroup = "consumer-group"
+	primaryRelationRoute         = "route"
+	primaryRelationService       = "service"
+)
+
 type stateBuilder struct {
 	targetContent   *Content
 	rawState        *utils.KongRawState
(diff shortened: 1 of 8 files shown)


pkg/file/builder.go
```diff
@@ -488,6 +488,10 @@ func (b *stateBuilder) consumers() {
 		// plugins for the Consumer
 		var plugins []FPlugin
 		for _, p := range c.Plugins {
+			if err := checkForNestedForeignKeys(p.Plugin, "consumer"); err != nil {
```
pmalek, line 491: Can we extract those magic strings into constants?

pmalek (commented):
Looks good. Just 1 minor nit. 

Can we also add the changelog entry with this PR?

Prashansa-K:
> Looks good. Just 1 minor nit.
> 
> Can we also add the changelog entry with this PR?

Thank you!
I will add a changelog in deck in the changelog.md file. Here, we add the changelog in release time, on the release page itself. Do I need to add it somewhere beforehand too?

pmalek (approved):

pmalek:
> Here, we add the changelog in release time, on the release page itself. Do I need to add it somewhere beforehand too?

Sorry. I get these mixed up as those repos have different approaches to handling changelog. 

👍

PR #159, opened 2025-01-08
fix: fixing consumer-group related dump config for deck.

Prashansa-K:
With changes added in go-kong, to skip
listing consumers while doing a GET call
on a consumer_group, dump config with gw
3.9+ starting showing issues. Associations
between a consumer and consumer-group were
missing. Thus, we are pushing the no listing
consumers feature behind a flag for now.

This change adds a new config that can be
set via a deck command flag. If set to true,
it will not list consumers-consumer group
associations.

Depends on https://github.com/Kong/go-kong/pull/494

Related deck PR: https://github.com/Kong/deck/pull/1486

### Issues resolved

For https://github.com/Kong/deck/issues/1483

### Documentation

### Testing

- [ ] Unit tests
- [ ] E2E tests


mheap:
@Prashansa-K This flag should apply to diff and sync too. It’s suitable when consumers are being managed in a different file.

By default, consumers should be included in all commands

Prashansa-K:
> @Prashansa-K This flag should apply to diff and sync too. It’s suitable when consumers are being managed in a different file.
> 
> By default, consumers should be included in all commands

Alright. That should be fine for this PR. We should not require any more code changes to support the said flag with sync and diff. I will update the description.

GGabriele (approved):
LGTM.

Remember: Konnect API doesn't support the `list_consumers` query param, so the flag you are planning to add in decK is not compatible when running against Konnect

PR #165, opened 2025-01-16
fix: avoid assuming api.konghq in URI and support private link global api endpoint

harshadixit12:
fix: avoid assuming api.konghq in URI and support private link global api endpoint

### Summary

IAM for konnect is global, and we call the global endpoint during login. 
This breaks for konnect over privatelink since the base URL does not contain the separator we are looking for.

How have we fixed it?
While creating the baseURL for global endpoint, we are checking for a new separator first for privatelink, since it is unlikely to change.

Why did we choose this approach?
We cannot use the Konnect address provided as-is since IAM is global, and need to switch region in subdomain to global

More details:
https://konghq.atlassian.net/wiki/spaces/~712020bb1b1f365f32471bbc7f249f5d9df3ef/pages/4300111923/Konnect+URL+invalid+when+api.konghq+is+not+in+passed+url

How have I tested this?
- [x] Manually ran the gateway ping command with different --konnect-addr flags and verified it works
- [ ] Shared binary with the reporter and tested

### Issues resolved

Fix #1488 in Deck
https://github.com/Kong/deck/issues/1488

### Testing

- [x] Unit tests
- [ ] E2E tests

### DIFF
--- a/pkg/konnect/login_service.go
+++ b/pkg/konnect/login_service.go
@@ -47,14 +47,19 @@ func (s *AuthService) Login(ctx context.Context, email,
 
 // getGlobalEndpoint returns the global endpoint for a given base Konnect URL.
 func getGlobalEndpoint(baseURL string) string {
+	const (
+		privateLinkRegionSeparator = "svc.konghq"
+		regionSeparator            = "api.konghq"
+	)
+
 	// If svc.konghq is present in string, continue with global endpoint private link
-	if strings.Contains(baseURL, "svc.konghq") {
-		parts := strings.Split(baseURL, "svc.konghq")
+	if strings.Contains(baseURL, privateLinkRegionSeparator) {
+		parts := strings.Split(baseURL, privateLinkRegionSeparator)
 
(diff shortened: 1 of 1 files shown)


Prashansa-K:
@harshadixit12 Let's add a unit test for the changes made in this PR.

harshadixit12:
> @harshadixit12 Let's add a unit test for the changes made in this PR.

@Prashansa-K I see the repo currently has integration tests - should I create new unit test and update the github pipeline workflow to run the unit test as well?

harshadixit12:
> > > @harshadixit12 Let's add a unit test for the changes made in this PR.
> > 
> > 
> > @Prashansa-K I see the repo currently has integration tests - should I create new unit test and update the github pipeline workflow to run the unit test as well?
> 
> @harshadixit12 There are unit tests too in this repo that run in the CI Test workflow (.github/workflows/test.yaml). For this change, the necessary tests should go to login_service_test.go file.

Ah right, my bad, was looking in the wrong dir for unit tests.

Prashansa-K (changes requested):

PR #178, opened 2025-01-24
fix: fixed fake custom entity panic issue

Prashansa-K:
### Summary

- In case a fake entity is passed to deck sync, a panic was received due to improper handling. Fixed that issue and added a test-case for the same.
- Renamed files to maintain proper numbering.


pkg/file/types.go
```diff
@@ -949,7 +949,7 @@ func (f *FCustomEntity) UnmarshalJSON(b []byte) error {
 		}
 		return copyToFCustomEntity(entity, f)
 	default:
-		return fmt.Errorf("unknown entity type: %s", *f.Type)
+		return fmt.Errorf("unknown entity type: %s", temp["type"])
```
harshadixit12, line 952: Out of curiosity, why would this panic before?

harshadixit12 (approved):

pkg/file/types.go
pmalek: Either `f` had to be nil or `.Type` (since it's dereferenced)

PR #182, opened 2025-01-27
Move `isPartialApply` out of deck in to GDR

mheap:
### Summary

Allow users to sync partial configuration to a running Gateway. If `isPartialApply` is set on `dumpConfig`, then all existing state is loaded in to `b.intermediate` to allow foreign key lookups to work. See https://github.com/Kong/deck/pull/1508/files#diff-7cf43e4538f1ca35fa6fbeb6fab1c4a8d084fd775b07133c86614b3462c9bbd9 for integration tests

### Full changelog

* Support `isPartialApply` in `dumpConfig` to enable incremental configuration changes
* Load all existing entities in to `b.intermediate` to enable foreign key lookups to work when applying partial configuration
* Add an `AddIgnoringDuplicates` method for use with `b.intermediate` to prevent duplicate state errors

### Testing

- [x] Unit tests
- [x] E2E tests (added via decK)


pkg/state/service_test.go
```diff
@@ -104,6 +104,19 @@ func TestServicesCollection_Add(t *testing.T) {
...
+	assert := assert.New(t)
+	collection := servicesCollection()
+
+	var s Service
+	s.ID = kong.String("my-id")
+	s.Name = kong.String("first")
+	err := collection.Add(s)
+	assert.Nil(err)
+	err = collection.AddIgnoringDuplicates(s)
+	assert.Nil(err)
```
pmalek, line 117: Food for thought. ( but as I can see that's currently done in all the tests added in this PR so ideally if this is changed we'd do it everywhere for consistency).

```suggestion
	collection := servicesCollection()

	var s Service
	s.ID = kong.String("my-id")
	s.Name = kong.String("first")
	require.NoError(t, collection.Add(s))
	require.NoError(t, collection.AddIgnoringDuplicates(s))
```

pkg/state/service.go
```diff
@@ -33,6 +33,30 @@ var serviceTableSchema = &memdb.TableSchema{
 // ServicesCollection stores and indexes Kong Services.
 type ServicesCollection collection
 
+func (k *ServicesCollection) AddIgnoringDuplicates(service Service) error {
```
pmalek, line 36: Would be nice to add a godoc for this `func` (actually there should be a linter catching that 🤔 ) explaining what it does and what's the use case.

Ditto for all the other `AddIgnoringDuplicates` funcs.

pkg/state/service.go
```diff
@@ -33,6 +33,30 @@ var serviceTableSchema = &memdb.TableSchema{
...
+	// Detect duplicates
+	if !utils.Empty(service.ID) {
```
pmalek, line 38: Is it possible for this to be called with `service`, `service.ID` or `service.Name` being `nil`? If so this will `panic`.

pkg/state/service.go
pmalek, line 1: Comment for possible future refactor: if all core entities like `Route`, `Service` and `Consumer` had methods for `Name() string` and `ID() string` then `AddIgnoringDuplicates` could be implemented just once based on an interface.

pkg/file/builder.go
```diff
@@ -175,6 +177,53 @@ func (b *stateBuilder) consumerGroups() {
...
+			}
+
+			var consumers []*kong.Consumer
+			for _, cgc := range consumerGroupConsumers {
+				if cgc.ConsumerGroup.ID == cg.ID {
+					consumers = append(consumers, cgc.Consumer)
+				}
+			}
+
+			rawCg.Consumers = consumers
```
pmalek, line 220: Move `rawCg` close to where it's used

```suggestion
			// Fetch consumers for this CG
			consumerGroupConsumers, err := b.currentState.ConsumerGroupConsumers.GetAll()
			if err != nil {
				b.err = err
				return
			}

			var consumers []*kong.Consumer
			for _, cgc := range consumerGroupConsumers {
				if cgc.ConsumerGroup.ID == cg.ID {
					consumers = append(consumers, cgc.Consumer)
				}
			}

			// Also add to rawState to perform a lookup later
			rawCg := &kong.ConsumerGroupObject{
				ConsumerGroup: &kong.ConsumerGroup{
					ID:        cg.ID,
					Name:      cg.Name,
					CreatedAt: cg.CreatedAt,
					Tags:      cg.Tags,
				},
				Consumers: consumers,
			}
```

pkg/dump/dump.go
```diff
@@ -52,6 +52,8 @@ type Config struct {
 
 	// IsConsumerGroupScopedPluginSupported
 	IsConsumerGroupScopedPluginSupported bool
+
+	IsPartialApply bool
```
pmalek, line 56: Might be worth adding a comment to this flag.

pkg/diff/diff.go
```diff
@@ -582,7 +582,7 @@ func generateDiffString(e crud.Event, isDelete bool, noMaskValues bool) (string,
 }
 
 // Solve generates a diff and walks the graph.
-func (sc *Syncer) Solve(ctx context.Context, parallelism int, dry bool, isJSONOut bool) (Stats,
+func (sc *Syncer) Solve(ctx context.Context, parallelism int, dry bool, isJSONOut bool, noDeletes bool) (Stats,
```
pmalek, line 585: If possible we'd make a similar change as in https://github.com/Kong/deck/commit/152977482599301b8559d3c7f70801bcb5222a3a to not add yet another `bool` parameter here.

pkg/state/service.go
mheap: `Add()` has a check, so it won't panic. I'll add a test case to demonstrate

pkg/state/service_test.go
mheap: Updated in all tests

pkg/file/builder.go
```diff
@@ -175,6 +177,53 @@ func (b *stateBuilder) consumerGroups() {
...
+			rawCg := &kong.ConsumerGroupObject{
+				ConsumerGroup: &kong.ConsumerGroup{
+					ID:        cg.ID,
+					Name:      cg.Name,
+					CreatedAt: cg.CreatedAt,
+					Tags:      cg.Tags,
+				},
+			}
+
+			rawCg.Consumers = consumers
```
pmalek, line 220:
```suggestion
			rawCg := &kong.ConsumerGroupObject{
				ConsumerGroup: &kong.ConsumerGroup{
					ID:        cg.ID,
					Name:      cg.Name,
					CreatedAt: cg.CreatedAt,
					Tags:      cg.Tags,
				},
				Consumers: consumers,
			}
```

pmalek (approved):

pkg/state/consumer_group_test.go
```diff
@@ -0,0 +1,41 @@
...
+	assert := assert.New(t)
+	collection := consumerGroupsCollection()
+
+	var cg ConsumerGroup
+	cg.ID = kong.String("my-id")
+	cg.Name = kong.String("first")
+	err := collection.Add(cg)
+	assert.Nil(err)
+	err = collection.AddIgnoringDuplicates(cg)
+	assert.Nil(err)
```
pmalek, line 40: I'm OK leaving this as is just bear in mind that `.NoError()` has slightly better error formatting when `err != nil`, same in all other places this is used this way.

pmalek (approved):

PR #197, opened 2025-02-07
chore: fix imported deck version to point to a released tag

czeslavo:
### Summary

https://github.com/Kong/go-database-reconciler/pull/182 changed deck version to `v1.43.1-0.20250128141633-787a0a819968` which is no longer available (most likely it was from a temporary branch @mheap had created). This PR fixes that as currently GDR doesn't compile.

### Testing

- [ ] Unit tests
- [ ] E2E tests


czeslavo:
Looks like there's a bug in deck's dependency that's causing a data race in `Test_LintPlain` test. Proposed a fix upstream: https://github.com/daveshanley/vacuum/pull/618.

Prashansa-K:
deck v1.44.2 is released. We can update the go-mod file to use that.

PR #214, opened 2025-03-07
select tag mismatch leads to replace

harshadixit12:
### Summary
In this PR, we are adding a check for entities which could be overwritten by mistake when ID is specified in the deck files, and tags are different from their original tags.
The cause for this is discussed [here](https://konghq.atlassian.net/wiki/spaces/~712020bb1b1f365f32471bbc7f249f5d9df3ef/pages/4451500059/deck+replaces+existing+entity+when+id+matches+but+select-tag+doesn+t).

This issue does not apply to the following entities, so changes have not been made:
* Those that don't use PUT (target, degraphql, etc)
* Those that are not supported by deck (licence)
* Those that don't support tags (rbac roles, etc)

### Full changelog
* Fixes entities being overwritten when ID is duplicated with different tags

### Issues resolved
https://github.com/Kong/deck/issues/1543

### Pending
- [x] Make the same changes for credential entities like basic auth

### Testing

- [x] Integration Tests

### Depends on 
Go Kong - https://github.com/Kong/go-kong/pull/526


pkg/types/license.go
```diff
@@ -93,6 +93,8 @@ type licenseDiffer struct {
 	kind crud.Kind
 
 	currentState, targetState *state.KongState
+
```
harshadixit12, line 96: Not necessary, will revert.

tests/integration/sync_test.go
```diff
@@ -7668,3 +7668,75 @@ func Test_Sync_CustomEntities_Fake(t *testing.T) {
 	require.Error(t, err)
 	assert.ErrorContains(t, err, "unknown entity type: fake-entity")
 }
+
+func Test_Sync_Avoid_Overwrite_On_Select_Tag_Mismatch_With_ID(t *testing.T) {
```
harshadixit12, line 7672: I think some changes are needed in this to run this test for the right versions

tests/integration/testdata/sync/038-avoid-entity-rewrite-with-id-on-select-tag-mismatch/ca-certificate-initial.yaml
```diff
@@ -0,0 +1,33 @@
+_format_version: "3.0"
+ca_certificates:
+- cert: |-
+    -----BEGIN CERTIFICATE-----
```
harshadixit12, line 4: I've generated this locally, and has validity up to ~2030 - beyond which the test will fail and this needs update.

pkg/types/aclgroup.go
```diff
@@ -159,6 +161,14 @@ func (d *aclGroupDiffer) createUpdateACLGroup(aclGroup *state.ACLGroup) (*crud.E
 	currentACLGroup, err := d.currentState.ACLGroups.Get(
 		*aclGroup.Consumer.ID, *aclGroup.ID)
 	if errors.Is(err, state.ErrNotFound) {
+		if aclGroup.ID != nil {
+			remoteACLGroup, _ := d.client.ACLs.GetByID(context.TODO(), aclGroup.ID)
```
harshadixit12, line 165: I've missed handling the errors here - 404 is expected, but others are not. 
I'll use `kong.IsNotFoundErr(err)` to fix this.

go.mod
```diff
@@ -21,7 +21,7 @@ require (
 	github.com/hashicorp/go-retryablehttp v0.7.7
 	github.com/hexops/gotextdiff v1.0.3
 	github.com/kong/deck v1.44.2
-	github.com/kong/go-kong v0.64.1
+	github.com/kong/go-kong v0.64.2-0.20250325074343-56c6154aef8d
```
harshadixit12, line 24: To be updated after go-kong is released

tests/integration/testdata/sync/039-avoid-entity-rewrite-with-id-on-select-tag-mismatch/acl-group-initial.yaml
```diff
@@ -0,0 +1,26 @@
...
+  - group: consumer's ACL
+    id: f2be545c-45c5-46e2-8acf-00ff2d899073
+    tags:
+    - before
+  custom_id: custom_id_1
+  id: 27d847fb-7aeb-4cb7-adb5-eccdd0ff1d42
+  username: consumer1
+  tags:
+  - before
+plugins:
```
Prashansa-K, line 13: Do we require plugin config for this test? We aren't testing for acl functionality anyway.
Does anything error out if it is not present?

Prashansa-K: Same for other cred types.

tests/integration/sync_test.go
```diff
@@ -7812,3 +7812,237 @@ func Test_Sync_Partials(t *testing.T) {
...
+			name:             "upstream",
+			initialStateFile: "testdata/sync/039-avoid-entity-rewrite-with-id-on-select-tag-mismatch/upstream-initial.yaml",
+			targetStateFile:  "testdata/sync/039-avoid-entity-rewrite-with-id-on-select-tag-mismatch/upstream-final.yaml",
+			errorExpected:    "error: an upstream with ID 87b6a97e-f3f7-4c47-857a-7464cb9e202b already exists",
+		},
+	}
+
+	for _, tc := range tests {
+		t.Run(tc.name, func(t *testing.T) {
+			runWhenKongOrKonnect(t, ">=2.8.0")
```
Prashansa-K, line 7922: This should be the first line of the test func. It ensures that we fetch the Kong client only for versions supported by the test.

tests/integration/sync_test.go
```diff
@@ -7812,3 +7812,237 @@ func Test_Sync_Partials(t *testing.T) {
 		require.NoError(t, err)
 	})
 }
+
+// test scope:
+//
+//	>=2.8.0 - for kong, konnect
```
Prashansa-K, line 7818:
```suggestion
// test scope:
//   - >=2.8.0
//   - konnect
```

Prashansa-K: Follow similar convention elsewhere

tests/integration/sync_test.go
```diff
@@ -7812,3 +7812,237 @@ func Test_Sync_Partials(t *testing.T) {
...
+func Test_Sync_Avoid_Overwrite_On_Select_Tag_Mismatch_With_ID(t *testing.T) {
+	setup(t)
+
+	// setup stage
+	client, err := getTestClient()
+	if err != nil {
```
Prashansa-K, line 7824: Let's use `require.NoError(t, err)` instead of this.

tests/integration/sync_test.go
```diff
@@ -7812,3 +7812,237 @@ func Test_Sync_Partials(t *testing.T) {
...
+
+	for _, tc := range tests {
+		t.Run(tc.name, func(t *testing.T) {
+			runWhenKongOrKonnect(t, ">=2.8.0")
+			mustResetKongState(context.TODO(), t, client, deckDump.Config{})
+			err := sync(tc.initialStateFile)
+			require.NoError(t, err)
+
+			err = sync(tc.targetStateFile, "--select-tag", "after")
+			require.ErrorContains(t, err, tc.errorExpected)
```
Prashansa-K, line 7928: We should "require" for the error here and then "assert" for the error message.
This would stop the test suite if no error comes, so as to avoid panics in next line.
If an assertion fails, the test suite would keep running, which is fine because we don't expect a panic there.

Stopping the entire test suit just because error strings didn't match doesn't make much sense.

```suggestion
			require.Error(t, err)
                        assert.ErrorContains(t, err, tc.errorExpected)
```

tests/integration/sync_test.go
Prashansa-K: Follow elsewhere too.

tests/integration/sync_test.go
```diff
@@ -7812,3 +7812,237 @@ func Test_Sync_Partials(t *testing.T) {
...
+			initialStateFile: "testdata/sync/039-avoid-entity-rewrite-with-id-on-select-tag-mismatch/upstream-initial.yaml",
+			targetStateFile:  "testdata/sync/039-avoid-entity-rewrite-with-id-on-select-tag-mismatch/upstream-final.yaml",
+			errorExpected:    "error: an upstream with ID 87b6a97e-f3f7-4c47-857a-7464cb9e202b already exists",
+		},
+	}
+
+	for _, tc := range tests {
+		t.Run(tc.name, func(t *testing.T) {
+			runWhenKongOrKonnect(t, ">=2.8.0")
+			mustResetKongState(context.TODO(), t, client, deckDump.Config{})
```
Prashansa-K, line 7923: Initialise a ctx at the start and pass it here. Follow elsewhere.

```
ctx := t.Context()
```
Follow elsewhere.

Prashansa-K:
There are a few stray newlines here and there. Let's clean them up too.

pkg/types/jwtauth.go
```diff
@@ -157,6 +159,18 @@ func (d *jwtAuthDiffer) createUpdateJWTAuth(jwtAuth *state.JWTAuth) (*crud.Event
...
 	if errors.Is(err, state.ErrNotFound) {
+		if jwtAuth.ID != nil {
+			existingJWTAuth, err := d.client.JWTAuths.GetByID(context.TODO(), jwtAuth.ID)
+
+			if err != nil && !kong.IsNotFoundErr(err) {
+				return nil, err
+			}
+
+			if existingJWTAuth != nil {
+				return nil, fmt.Errorf("error: a jwt credential with ID %s already exists", *jwtAuth.ID)
```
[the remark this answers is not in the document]
harshadixit12: Makes sense, it'll be easier to ensure consistency in the error message as well.

tests/integration/testdata/sync/038-avoid-entity-rewrite-with-id-on-select-tag-mismatch/ca-certificate-initial.yaml
Prashansa-K: Can you add a comment in the file to state the same? It will be easier to track that way.

tests/integration/testdata/sync/039-avoid-entity-rewrite-with-id-on-select-tag-mismatch/acl-group-initial.yaml
harshadixit12: Yep, plugin was not needed - i think I missed it because it was needed on UI

tests/integration/sync_test.go
harshadixit12: I'm a bit confused here - we want to use require.Error here as good practice, correct? (here panic is not happening, but it could in other tests if we don't use?)
Also didn't understand `Stopping the entire test suit just because error strings didn't match doesn't make much sense.` - here the rest of the subtests run as expected when when one error string doesn't match, even when require.Error is not used - what am I missing? 🤔 
@Prashansa-K

Prashansa-K (approved):

Issue #166, opened 2025-01-21
Partials support


mheap:
Support for the new `partials` entity in Kong Gateway

Internal design doc: https://docs.google.com/document/d/1vwUS5aCHXOjE5jYaysU8eJ59da5OdskoNd58pJGLO2g/edit?tab=t.0#heading=h.20vb8eyhpj3d

This support needs adding in GDR and decK

PR #215, opened 2025-03-11
feat: added support for partials and plugin-partial association

Prashansa-K:
### Summary

SUMMARY_GOES_HERE

### Full changelog

* [Implement ...]
* [Fix ...]

### Issues resolved

For https://github.com/Kong/go-database-reconciler/issues/166

### Documentation

### Testing

- [ ] Unit tests
- [ ] E2E tests


go.mod
```diff
@@ -14,22 +14,22 @@ require (
...
-	github.com/google/go-cmp v0.6.0
+	github.com/google/go-cmp v0.7.0
 	github.com/google/go-querystring v1.1.0
 	github.com/google/uuid v1.6.0
 	github.com/hashicorp/go-memdb v1.3.4
 	github.com/hashicorp/go-retryablehttp v0.7.7
 	github.com/hexops/gotextdiff v1.0.3
 	github.com/kong/deck v1.44.2
-	github.com/kong/go-kong v0.63.0
+	github.com/kong/go-kong v0.63.1-0.20250311100117-3fb9f0059d0c
```
Prashansa-K, line 24: This will be updated once go-kong is released.
https://github.com/Kong/go-kong/pull/507

Prashansa-K:
Hi @pmalek 
Would you be able to help with this PR review? I assigned this to you as you also reviewed the related go-kong PR.
Thanks again for that!

pkg/diff/diff.go
```diff
@@ -625,6 +627,26 @@ func (sc *Syncer) Solve(ctx context.Context, parallelism int, dry bool, isJSONOu
 		// configuration that is sent to Kong.
 		eventForKong := e
 
+		findLinkedPartials := func(plugin *kong.Plugin) ([]*kong.Partial, error) {
```
pmalek, line 630: Could we make this a regular func (and ideally add a UT for it)?

pkg/file/builder.go
```diff
@@ -182,6 +183,44 @@ func (b *stateBuilder) addConsumerGroupPlugins(
...
+			partial, err = b.currentState.Partials.Get(*p.Name)
+		}
+		if utils.Empty(p.ID) {
+			if errors.Is(err, state.ErrNotFound) || partial == nil {
+				p.ID = uuid()
+			} else if err != nil {
+				b.err = err
+				return
+			} else {
+				p.ID = kong.String(*partial.ID)
```
pmalek, line 206: Can `partial.ID` be `nil` here? That would cause a panic.

As far as I can see this could happen when `utils.Empty(p.Name)` returns `true` right?

pkg/file/builder.go
```diff
@@ -1421,6 +1460,40 @@ func (b *stateBuilder) plugins() {
 			p.ConsumerGroup = utils.GetConsumerGroupReference(cg.ConsumerGroup)
 		}
 
+		if p.Partials != nil {
+			var pluginPartials []*kong.PartialLink
+			for _, partial := range p.Partials {
+				if partial.Partial != nil {
```
pmalek, line 1466: To prevent deep nesting can we revert the condition here to `if partial.Partial == nil` and `continue` if that happens?

tests/integration/diff_test.go
```diff
@@ -2944,3 +2944,35 @@ func Test_Diff_NoDeletes_3x(t *testing.T) {
...
+	syncer, err := deckDiff.NewSyncer(deckDiff.SyncerOpts{
+		CurrentState: currentState,
+		TargetState:  targetState,
+
+		KongClient: client,
+	})
+	require.NoError(t, err)
+
+	stats, errs, changes := syncer.Solve(ctx, 1, true, true)
+	require.Len(t, errs, 0, "Should have no errors in diffing")
```
pmalek, line 2972:
```suggestion
	require.Empty(t, errs, "Should have no errors in diffing")
```

tests/integration/sync_test.go
```diff
@@ -7668,3 +7668,166 @@ func Test_Sync_CustomEntities_Fake(t *testing.T) {
...
+		syncer, err := deckDiff.NewSyncer(deckDiff.SyncerOpts{
+			CurrentState: currentState,
+			TargetState:  targetState,
+
+			KongClient: client,
+		})
+		require.NoError(t, err)
+
+		stats, errs, changes := syncer.Solve(ctx, 1, false, true)
+		require.Len(t, errs, 0, "Should have no errors in syncing")
```
pmalek, line 7722:
```suggestion
		require.Empty(t, errs, "Should have no errors in syncing")
```

tests/integration/sync_test.go
```diff
@@ -7668,3 +7668,166 @@ func Test_Sync_CustomEntities_Fake(t *testing.T) {
 	require.Error(t, err)
 	assert.ErrorContains(t, err, "unknown entity type: fake-entity")
 }
+
+func Test_Sync_Partials_Plugins(t *testing.T) {
+	client, err := getTestClient()
+	if err != nil {
+		t.Fatalf(err.Error())
+	}
```
pmalek, line 7676:
```suggestion
	require.NoError(t, err)
```

tests/integration/sync_test.go
```diff
@@ -7668,3 +7668,166 @@ func Test_Sync_CustomEntities_Fake(t *testing.T) {
...
+		require.Error(t, err)
+		assert.ErrorContains(t, err, "partial non-existent-partial for plugin rate-limiting-advanced: entity not found")
+	})
+}
+
+func Test_Sync_Partials(t *testing.T) {
+	client, err := getTestClient()
+	if err != nil {
+		t.Fatalf(err.Error())
+	}
```
pmalek, line 7808:
```suggestion
	require.NoError(t, err)
```

tests/integration/sync_test.go
```diff
@@ -7668,3 +7668,166 @@ func Test_Sync_CustomEntities_Fake(t *testing.T) {
...
+		syncer, err := deckDiff.NewSyncer(deckDiff.SyncerOpts{
+			CurrentState: currentState,
+			TargetState:  targetState,
+
+			KongClient: client,
+		})
+		require.NoError(t, err)
+
+		stats, errs, changes := syncer.Solve(ctx, 1, false, true)
+		require.Len(t, errs, 0, "Should have no errors in syncing")
```
pmalek, line 7766:
```suggestion
		require.Empty(t, errs, "Should have no errors in syncing")
```

pkg/types/postProcess.go
```diff
@@ -528,3 +528,20 @@ func (crud *degraphqlRoutePostAction) Delete(_ context.Context, args ...crud.Arg
...
+type partialPostAction struct {
+	currentState *state.KongState
+}
+
+func (crud partialPostAction) Create(_ context.Context, args ...crud.Arg) (crud.Arg, error) {
+	return nil, crud.currentState.Partials.Add(*args[0].(*state.Partial))
+}
+
+func (crud partialPostAction) Delete(_ context.Context, args ...crud.Arg) (crud.Arg, error) {
+	// TODO: add check for linked plugins prior to deletion
```
pmalek, line 541: Do we have an issue for this to link here?

pkg/state/types.go
```diff
@@ -1842,3 +1842,50 @@ func (d *DegraphqlRoute) EqualWithOpts(d2 *DegraphqlRoute, ignoreID bool) bool {
...
+
+// Equal returns true if partials p and p2 are equal.
+func (p *Partial) Equal(p2 *Partial) bool {
+	return p.EqualWithOpts(p2, false, false)
+}
+
+// EqualWithOpts returns true if partials p and p2 are equal.
+// If ignoreID is set to true, IDs will be ignored while comparison.
+// If ignoreTS is set to true, timestamp fields will be ignored.
+func (p *Partial) EqualWithOpts(p2 *Partial, ignoreID, ignoreTS bool) bool {
```
pmalek, line 1875: This can be a bit hard to reason about at call sites having 2 bool arguments.

WDYT about making this access functional options and providing e.g. `IgnoreID()` and `IgnoreTS()` option funcs which would act on `*Partial` (clearing `.ID` and timestamps respectively)? This would not only make this clearer to reason about at call site but also possible to be used with different sets of these options (e.g. ignoring only timestamps).

pkg/state/partials_test.go
```diff
@@ -0,0 +1,303 @@
...
+func TestPartialGetAll(t *testing.T) {
+	assert := assert.New(t)
+	require := require.New(t)
+	collection := partialsCollection()
+
+	populatePartials(t, collection)
+
+	partials, err := collection.GetAll()
+	require.NoError(err, "error getting all partials")
+	assert.Equal(5, len(partials))
```
pmalek, line 238:
```suggestion
	assert.Len(partials, 5)
```

pkg/types/postProcess.go
Prashansa-K: This is already covered in the gateway now and we don't need to add separate checks anymore.

pkg/file/builder.go
Prashansa-K: I have added a check on L200 to see if partial is nil. If it is, we generate an id.
Added another one to check for the ID as well. It is very less likely to happen but better to catch it than panic.

pkg/diff/diff.go
Prashansa-K: I have added a separate function for the same in utils package.
Not sure how to add a UT for this as it requires a kongClient. I would have to yank and put the same logic we use in integration tests for fetching a client. That approach didn't seem well.

Is there any other suggested way? 
I am not a fan for mocking the client. I find that approach not so useful because it's very self-fulfilling. The mock may return the necessary data as we need (as we are the ones adding the mock). So, it can't help in case the actual kong client has started returning different data/no data or stuff like that. So, it doesn't fail!

pkg/state/types.go
Prashansa-K: I agree but I would like to keep it this way, so that it adheres to the formats other entities use for the same operation. If we need to change it, we should do it in all places to make things uniform. This can be tracked as a separate improvement.

tests/integration/sync_test.go
```diff
@@ -7652,3 +7652,162 @@ func Test_Sync_CustomEntities_Fake(t *testing.T) {
...
+}
+
+func Test_Sync_Partials(t *testing.T) {
+	client, err := getTestClient()
+	require.NoError(t, err)
+
+	ctx := context.Background()
+
+	runWhen(t, "enterprise", ">=3.10.0")
+	dumpConfig := deckDump.Config{}
```
pmalek, line 7793:
```suggestion
	runWhen(t, "enterprise", ">=3.10.0")

	client, err := getTestClient()
	require.NoError(t, err)

	ctx := t.Context()

	dumpConfig := deckDump.Config{}
```

tests/integration/sync_test.go
```diff
@@ -7652,3 +7652,162 @@ func Test_Sync_CustomEntities_Fake(t *testing.T) {
...
+
+		newState, err := fetchCurrentState(ctx, client, dumpConfig)
+		require.NoError(t, err)
+
+		// check for partial
+		partials, err := newState.Partials.GetAll()
+		require.NoError(t, err)
+		require.NotNil(t, partials)
+
+		assert.Len(t, partials, 1)
```
pmalek, line 7759: All of `Len` asserts should use `require` to prevents panics in subsequent lines

```suggestion
		require.Len(t, partials, 1)
```

tests/integration/sync_test.go
```diff
@@ -7652,3 +7652,162 @@ func Test_Sync_CustomEntities_Fake(t *testing.T) {
...
 }
+
+func Test_Sync_Partials_Plugins(t *testing.T) {
+	client, err := getTestClient()
+	require.NoError(t, err)
+
+	ctx := context.Background()
+
+	runWhen(t, "enterprise", ">=3.10.0")
+	dumpConfig := deckDump.Config{}
```
pmalek, line 7663: No need to create client if Kong version is not as expected.

```suggestion
	runWhen(t, "enterprise", ">=3.10.0")

	client, err := getTestClient()
	require.NoError(t, err)

	ctx := t.Context()

	dumpConfig := deckDump.Config{}
```

tests/integration/diff_test.go
```diff
@@ -2945,3 +2945,35 @@ func Test_Diff_NoDeletes_3x(t *testing.T) {
 		})
 	}
 }
+
+func Test_Diff_Partials(t *testing.T) {
+	runWhen(t, "enterprise", ">=3.10.0")
+	client, err := getTestClient()
+	if err != nil {
+		t.Fatalf(err.Error())
+	}
```
pmalek, line 2954:
```suggestion
	require.NoError(t, err)
```

pkg/state/types.go
pmalek: Fair enough, do you mind creating an issue for this?

pkg/state/partials.go
```diff
@@ -0,0 +1,207 @@
...
+	iter, err := txn.Get(partialTableName, all, true)
+	if err != nil {
+		return nil, err
+	}
+
+	var res []*Partial
+	for el := iter.Next(); el != nil; el = iter.Next() {
+		p, ok := el.(*Partial)
+		if !ok {
+			panic(unexpectedType)
```
pmalek, line 201: Shouldn't we return an error here instead of panicking?

pkg/state/partials.go
```diff
@@ -0,0 +1,207 @@
...
+		}
+		return &Partial{Partial: *partial.DeepCopy()}, nil
+	}
+	return nil, ErrNotFound
+}
+
+// Add adds a partial to the collection.
+// partial.ID should not be nil else an error is thrown.
```
pmalek, line 57:
```suggestion
// partial.ID should not be nil or an error is thrown.
```

pkg/state/partials.go
```diff
@@ -0,0 +1,207 @@
...
+		if errors.Is(err, ErrNotFound) {
+			continue
+		}
+		if err != nil {
+			return nil, err
+		}
+
+		partial, ok := res.(*Partial)
+		if !ok {
+			panic(unexpectedType)
```
pmalek, line 49: Ditto: shouldn't we return an error instead?

pkg/state/types.go
Prashansa-K: Done.
https://github.com/Kong/go-database-reconciler/issues/223

pkg/diff/diff.go
```diff
@@ -660,13 +667,44 @@ func (sc *Syncer) Solve(ctx context.Context, parallelism int, dry bool, isJSONOu
...
+		// If the event is for a partial, inject defaults in the partial's config
+		// that will be used for the diff. This is needed to avoid highlighting
+		// default values that were populated by Kong as differences.
+		if partial, ok := e.Obj.(*state.Partial); ok {
+			partialCopy := &state.Partial{Partial: *partial.DeepCopy()}
+			e.Obj = partialCopy
+
+			if exists, _ := utils.WorkspaceExists(ctx, sc.kongClient); exists {
+				var schema map[string]interface{}
+				schema, err = sc.kongClient.Partials.GetFullSchema(ctx, partialCopy.Partial.Type)
```
pmalek, line 696: Any reason to reuse the `err`? AFAICT it's only used in the line below. Line 701 already declares its own `err`.

```suggestion
				schema, err := sc.kongClient.Partials.GetFullSchema(ctx, partialCopy.Partial.Type)
```

pkg/dump/dump.go
```diff
@@ -487,6 +487,21 @@ func getProxyConfiguration(ctx context.Context, group *errgroup.Group,
...
 
+	group.Go(func() error {
+		partials, err := GetAllPartials(ctx, client, config.SelectorTags)
+		if err != nil {
+			return fmt.Errorf("partials: %w", err)
+		}
+
+		partials, err = excludeKonnectManagedEntities(partials)
+		if err != nil {
+			return fmt.Errorf("partials: %w", err)
```
pmalek, line 498: Just to differentiate these 2 errors:

```suggestion
			return fmt.Errorf("failed excluding Konnect managed entities from partials: %w", err)
```

tests/integration/sync_test.go
```diff
@@ -7652,3 +7652,163 @@ func Test_Sync_CustomEntities_Fake(t *testing.T) {
...
+	t.Run("create a partial and link to a plugin via name", func(t *testing.T) {
+		mustResetKongState(ctx, t, client, dumpConfig)
+		currentState, err := fetchCurrentState(ctx, client, dumpConfig)
+		require.NoError(t, err)
+
+		targetState := stateFromFile(ctx, t, "testdata/sync/038-partials/kong.yaml", client, dumpConfig)
+		syncer, err := deckDiff.NewSyncer(deckDiff.SyncerOpts{
+			CurrentState: currentState,
+			TargetState:  targetState,
+
```
pmalek, line 7699:
```suggestion
```

pmalek (approved):
Just minor nits. Otherwise, this looks good.

PR #88, opened 2024-05-08
fix: correct tags filtering on consumer-group consumers

GGabriele:
Due to the nature of the Consumer Group API, it's not possible to natively filter its Consumers using tags:

```
$ http :8001/consumer_groups/foo_group_1
HTTP/1.1 200 OK

{
    "consumer_group": {
        "created_at": 1715183554,
        "id": "e0e495f2-9acd-41f7-b4da-33659033af13",
        "name": "foo_group_1",
        "tags": [
            "managed-by-deck"
        ],
        "updated_at": 1715183554
    },
    "consumers": [
        {
            "created_at": 1715183563,
            "id": "a704b13e-1a9b-409b-9611-fd2bfcea63bb",
            "type": 0,
            "updated_at": 1715183563,
            "username": "foo",
            "username_lower": "foo"
        }
    ]
}
```

This commit makes sure the correct filtering is "manually" done, avoiding existing Consumers to be wiped out during syncs when using Tags.

### Summary

SUMMARY_GOES_HERE

### Full changelog

* [Implement ...]
* [Fix ...]

### Issues resolved

Fix #XXX

### Documentation

- [ ] Link to the website [documentation PR](https://github.com/Kong/docs.konghq.com/pull/XXX)

### Testing

- [ ] Unit tests
- [ ] E2E tests
- [ ] Manual testing on Universal
- [ ] Manual testing on Kubernetes

PR #228, opened 2025-03-26
fix: limiting tag-based lookups for consumer_group_consumers to only select-tags

Prashansa-K:
### Summary

While creating the dump-config, select-tags or 
defaultLookupTags are passed for all entities.
In the case of consumer-groups, tags are also
used to filter out the consumers inside the group.
This is useful for select-tags, as consumers may be
maintained outside of deck and we don't wish to
delete them. More information here: https://github.com/Kong/go-database-reconciler/pull/88
However, in case of defaultLookupTags, this filtering
leaves out consumers in case no tags match. 
This causes the dumpConfig to not consider some 
`consumer_group_consumers`, leading to another 
create event during a subsequent sync.
Since the gateway already has the consumer_group_consumer,
the create event leads to a failure.

Consider the below files and sync them in order as stated below.
```
# consumer-group.yaml
_format_version: "3.0"
consumer_groups:
  - name: foo-group
    tags:
      - group-tag
```

```
# consumers.yaml
_format_version: "3.0"
_info:
  defaults: {}
  select_tags:
    -  user-tag
  default_lookup_tags:
    consumer_groups:
      - group-tag

consumers:
  - username: user-1
    groups:
      - name: foo-group
    tags:
    - user-tag
```

`$ deck gateway sync consumer-group.yaml`
_creating consumer-group foo-group
Summary:
  Created: 1
  Updated: 0
  Deleted: 0_

`$ deck gateway sync consumers.yaml`
_creating consumer user-1
creating consumer-group-consumer user-1
Summary:
  Created: 2
  Updated: 0
  Deleted: 0_

`$ deck gateway sync consumers.yaml`
_creating consumer-group-consumer user-1
Summary:
  Created: 0
  Updated: 0
  Deleted: 0
Error: 1 errors occurred:
        while processing event: Create consumer-group-consumer user-1 failed: HTTP status 409 (message: "Consumer '9a1b6efc-45ef-4b09-8c3d-f391d288d700' already in group 'd6a1ba7f-2213-49be-996d-2584ba54de1e'")_

The second sync for `consumers.yaml` file leads to a failure event as shown above.

In this PR, we are passing a tag-type to the functions that
fetch consumer-groups. In case the tag is a `select-tag`,
we can continue the tag-based filtering for consumers.
Otherwise, we skip it.

With this change, we will ensure the necessary filtering
happens in case of select-tags, but is skipped while using
a defaultLookupTag.

### Issues resolved

For https://github.com/Kong/deck/issues/1544

### Documentation

### Testing

- [ ] Unit tests
- [x] E2E tests


Prashansa-K:
@pmalek I have changed the function signature for an exported function here. I understand that it's not a good practice but we already have four versions for `GetConsumerGroups` function. I checked that gdr module is used in deck and kubernetes-ingress-controller. In the latter, there's no direct call to any of the functions that are modified. So, I proceeded with this change thinking it to be a safe bet.
Let me know if that's okay.

tests/integration/sync_test.go
```diff
@@ -7821,3 +7821,45 @@ func Test_Sync_Partials(t *testing.T) {
...
 	})
 }
+
+func Test_Sync_Consumers_Default_Lookup_Tag(t *testing.T) {
+	runWhen(t, "enterprise", ">=3.0.0")
+
+	client, err := getTestClient()
+	require.NoError(t, err)
+
+	ctx := context.Background()
```
pmalek, line 7831:
```suggestion
	ctx := t.Context()
```

pkg/dump/dump.go
```diff
@@ -148,15 +155,16 @@ func getConsumerGroupsConfiguration(ctx context.Context, group *errgroup.Group,
 			getConsumerGroupsFunc = GetAllConsumerGroupsWithoutConsumers
 		}
 
-		consumerGroups, err = getConsumerGroupsFunc(ctx, client, config.SelectorTags)
+		consumerGroups, err = getConsumerGroupsFunc(ctx, client, config.SelectorTags, SelectTag)
```
pmalek, line 158: Can we add a comment here explaining the usage of the select tag and what are its effects?

pkg/dump/dump.go
```diff
@@ -148,15 +155,16 @@ func getConsumerGroupsConfiguration(ctx context.Context, group *errgroup.Group,
...
 		if err != nil {
 			if kong.IsNotFoundErr(err) || kong.IsForbiddenErr(err) {
 				return nil
 			}
 			return fmt.Errorf("consumer_groups: %w", err)
 		}
 		if config.LookUpSelectorTagsConsumerGroups != nil {
-			globalConsumerGroups, err := getConsumerGroupsFunc(ctx, client, config.LookUpSelectorTagsConsumerGroups)
+			globalConsumerGroups, err := getConsumerGroupsFunc(ctx, client, config.LookUpSelectorTagsConsumerGroups,
+				DefaultLookupTag)
```
pmalek, line 167: Ditto about the default tag.

pmalek:
> @pmalek I have changed the function signature for an exported function here. I understand that it's not a good practice but we already have four versions for `GetConsumerGroups` function. I checked that gdr module is used in deck and kubernetes-ingress-controller. In the latter, there's no direct call to any of the functions that are modified. So, I proceeded with this change thinking it to be a safe bet. Let me know if that's okay.

Understood. I think that it'd be best to use an `interface` somewhere in there instead assigning a func to a variable and unnecessarily changing these functions' declarations but I also understand that we don't want to refactor this as part of this fix.

harshadixit12 (approved):

PR #240, opened 2025-04-03
fix: ip restriction plugin diff

harshadixit12:
### Summary
The issue seems to be due to the order of `protocols` defined for the ip-restriction plugin.

When we sync for the second time, if the protocols are not declared in the file we are trying to sync, we load defaults from the `/schemas/plugins/ip-restriction` endpoint on konnect - which returns them in a [given order](https://github.com/Kong/kong-ee/blob/c31e7045db6a111c254e6609c1945e9320f6d661/kong/plugins/ip-restriction/schema.lua#L14).

However, when we fetch the current state, konnect returns the plugin with the protocols sorted.
This is leading to the differ considering them unequal.

Relevant discussion: https://kongstrong.slack.com/archives/CQK8J4VN3/p1743588627387659
 

### Full changelog

* [fix: ip restriction plugin diff](https://github.com/Kong/go-database-reconciler/pull/240/commits/c006d430e75c12537b324d162d9db29431e824ff)

### Issues resolved
https://github.com/Kong/deck/issues/1571

### Testing

- [x] Unit tests
- [x] E2E tests

### Pending
- [x] Add tests if relevant
- [x] Test with other plugins
- [x] Find out why only ip-restriction plugin schema has unsorted order and others don't
- [x] Do a quick search if any other fields can have same issue

### DIFF
--- a/pkg/state/types.go
+++ b/pkg/state/types.go
@@ -579,6 +579,10 @@ func (p1 *Plugin) EqualWithOpts(p2 *Plugin, ignoreID,
 	sort.Slice(p1Copy.Tags, func(i, j int) bool { return *(p1Copy.Tags[i]) < *(p1Copy.Tags[j]) })
 	sort.Slice(p2Copy.Tags, func(i, j int) bool { return *(p2Copy.Tags[i]) < *(p2Copy.Tags[j]) })
 
+	// sort the protocols
+	sort.Slice(p1Copy.Protocols, func(i, j int) bool { return *(p1Copy.Protocols[i]) < *(p1Copy.Protocols[j]) })
+	sort.Slice(p2Copy.Protocols, func(i, j int) bool { return *(p2Copy.Protocols[i]) < *(p2Copy.Protocols[j]) })
+
 	p1Copy.Config = sortNestedArrays(p1Copy.Config)
 	p2Copy.Config = sortNestedArrays(p2Copy.Config)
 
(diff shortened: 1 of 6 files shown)


pkg/state/types.go
```diff
@@ -579,6 +579,10 @@ func (p1 *Plugin) EqualWithOpts(p2 *Plugin, ignoreID,
 	sort.Slice(p1Copy.Tags, func(i, j int) bool { return *(p1Copy.Tags[i]) < *(p1Copy.Tags[j]) })
 	sort.Slice(p2Copy.Tags, func(i, j int) bool { return *(p2Copy.Tags[i]) < *(p2Copy.Tags[j]) })
 
+	// sort the protocols
```
Prashansa-K, line 582: Nit: Unnecessary comment.

pkg/state/types.go
```diff
@@ -579,6 +579,10 @@ func (p1 *Plugin) EqualWithOpts(p2 *Plugin, ignoreID,
...
+	sort.Slice(p1Copy.Protocols, func(i, j int) bool { return *(p1Copy.Protocols[i]) < *(p1Copy.Protocols[j]) })
```
Prashansa-K, line 583: Protocols are never nil, correct?

Prashansa-K (approved):

pkg/state/types.go
harshadixit12: I'll verify and confirm

harshadixit12: sort.Slice is safe to use on empty / nil slices (verified by running in go playground)

Prashansa-K (approved):

PR #248, opened 2025-04-09
chore: bump go-kong to v0.65.1

harshadixit12:
### Summary

This bumps go-kong version containing support for loading all pages of endpoint permissions for RBAC roles.

Go kong PR: https://github.com/Kong/go-kong/pull/533

### Issues resolved
https://github.com/Kong/deck/issues/1574

### Documentation

### Testing

- [ ] Unit tests
- [ ] E2E tests

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -21,7 +21,7 @@ require (
 	github.com/hashicorp/go-retryablehttp v0.7.7
 	github.com/hexops/gotextdiff v1.0.3
 	github.com/kong/deck v1.45.1-0.20250326064911-1a3c87c5bf97
-	github.com/kong/go-kong v0.65.0
+	github.com/kong/go-kong v0.65.1
 	github.com/samber/lo v1.49.1
 	github.com/shirou/gopsutil/v3 v3.24.5
 	github.com/ssgelm/cookiejarparser v1.0.1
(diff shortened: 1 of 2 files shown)


harshadixit12:
@Prashansa-K I've not added integration tests for 2 reasons - 
1. It was a bug in the dependency (go-kong), and unit test has been added there to test all pages were loaded.
2. To test this effectively at GDR / deck, we'd need to sync > 1000 rbac endpoint permissions for a role, and then dump and compare - this would take a few seconds to run for each version of gateway , and would have a bulky state file for this test.

Please let me know if you think adding integration test in deck / GDR would be required

Prashansa-K:
> @Prashansa-K I've not added integration tests for 2 reasons -
> 
> 1. It was a bug in the dependency (go-kong), and unit test has been added there to test all pages were loaded. (I'd want to trust the interface given by the dependency)
> 2. To test this effectively at GDR / deck, we'd need to sync > 1000 rbac endpoint permissions for a role, and then dump and compare - this would take a few seconds to run for each version of gateway , and would have a bulky state file for this test.
> 
> Please let me know if you think adding integration test in deck / GDR would be required

We don't need a test for this. We don't have it for other entities too.

Prashansa-K (approved):

PR #250, opened 2025-04-09
fix: named foreign key references work properly for plugins

Prashansa-K:
Till deck v1.42.1, references in plugin
config worked fine if IDs were passed.
If names were passed, the first sync
went fine, but all the next syncs
showed unique key violation error.
This was because names were getting
resolved as IDs and then processed for
fetching the plugin from currentState,
which gave a 404. Thus, a new plugin
was getting created each time, leading
to uniqueness violation error.

With deck v1.43.0, we were not
allowing nested foreign keys. We are 
relaxing this constraint here:
https://github.com/Kong/go-database-reconciler/pull/249
Thus, necessary changes are added 
to make this work properly with named references.

### Issues resolved

For https://github.com/Kong/deck/issues/1595

### Documentation

### Testing

- [x] Unit tests
- [x] E2E tests

### DIFF
--- a/pkg/file/builder.go
+++ b/pkg/file/builder.go
@@ -1685,9 +1685,9 @@ func (b *stateBuilder) ingestRoute(r FRoute) error {
 	// plugins for the route
 	var plugins []FPlugin
 	for _, p := range r.Plugins {
-		if err := checkForNestedForeignKeys(p.Plugin, primaryRelationRoute); err != nil {
-			return err
-		}
+		// if err := checkForNestedForeignKeys(p.Plugin, primaryRelationRoute); err != nil {
+		// 	return err
+		// }
 		p.Route = utils.GetRouteReference(r.Route)
 		plugins = append(plugins, *p)
 	}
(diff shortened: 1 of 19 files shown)


pkg/file/builder.go
```diff
@@ -1755,20 +1755,50 @@ func (b *stateBuilder) fillPluginConfig(plugin *FPlugin) error {
 	return nil
 }
 
-func pluginRelations(plugin *kong.Plugin) (cID, rID, sID, cgID string) {
+func (b *stateBuilder) pluginRelations(plugin *kong.Plugin) (cID, rID, sID, cgID string) {
 	if plugin.Consumer != nil && !utils.Empty(plugin.Consumer.ID) {
-		cID = *plugin.Consumer.ID
+		consumer, err := b.currentState.Consumers.GetByIDOrUsername(*plugin.Consumer.ID)
```
harshadixit12, line 1760: We're making this change so actual ID is returned instead of name when foreign key reference exists, correct?

Prashansa-K: Yes!
The internal state that we maintain uses IDs. So, if a name is passed to it, even if the said entity exists, the lookup happens via the name and fails with a 404.

harshadixit12: https://github.com/Kong/go-database-reconciler/blob/63d97250ecf59707d722d0646b1e0d6912bc9d00/pkg/file/builder.go#L1711 is the lookup that needs the ID - and returns nil if name is given - leading to unique violation eventually as we try to create instead of update - correct?

Prashansa-K: Yes. Correct!

PR #267, opened 2025-05-06
fix: partial apply for custom entities

harshadixit12:
### Summary

In this PR, we are ignoring the error returned from `ensureService` for Degraphql routes - similar to how other entities behave so that gateway apply command is fixed for custom entities.

### Issues resolved
https://github.com/Kong/deck/issues/1568

### Documentation

### Testing

- [ ] Unit tests
- [x] E2E tests

### DIFF
--- a/pkg/state/builder.go
+++ b/pkg/state/builder.go
@@ -405,10 +405,9 @@ func buildKong(kongState *KongState, raw *utils.KongRawState) error {
 			if err != nil {
 				return err
 			}
-			if !ok {
-				return fmt.Errorf("service %q not found", *d.Service.ID)
+			if ok {
+				d.Service = s
 			}
-			d.Service = s
 		}
 		err := kongState.DegraphqlRoutes.Add(DegraphqlRoute{DegraphqlRoute: *d})
 		if err != nil {
--- a/tests/integration/apply_test.go
+++ b/tests/integration/apply_test.go
@@ -1,0 +1,43 @@
+//go:build integration
+
+package integration
+
+import (
+	"context"
+	"testing"
+
+	deckDump "github.com/kong/go-database-reconciler/pkg/dump"
+	"github.com/stretchr/testify/require"
+)
+
+func Test_Apply_Custom_Entities(t *testing.T) {
+	runWhen(t, "enterprise", ">=3.0.0")
+	setup(t)
+	client, err := getTestClient()
+	if err != nil {
+		t.Fatalf(err.Error())
+	}
+	ctx := context.Background()
+	tests := []struct {
+		name                   string
+		initialStateFile       string
+		targetPartialStateFile string
+	}{
+		{
+			name:                   "certificate",
+			initialStateFile:       "testdata/apply/001-custom-entities/initial-state.yaml",
+			targetPartialStateFile: "testdata/apply/001-custom-entities/partial-update.yaml",
+		},
+	}
+
+	for _, tc := range tests {
+		t.Run(tc.name, func(t *testing.T) {
+			mustResetKongState(ctx, t, client, deckDump.Config{})
+			err := sync(tc.initialStateFile)
+			require.NoError(t, err)
+
+			err = apply(tc.targetPartialStateFile)
+			require.NoError(t, err)
+		})
+	}
+}
--- a/tests/integration/apply_test.go
+++ b/tests/integration/apply_test.go
@@ -24,7 +24,7 @@ func Test_Apply_Custom_Entities(t *testing.T) {
 		targetPartialStateFile string
 	}{
 		{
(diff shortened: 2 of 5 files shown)


pkg/state/builder.go
```diff
@@ -405,10 +405,9 @@ func buildKong(kongState *KongState, raw *utils.KongRawState) error {
 			if err != nil {
 				return err
 			}
-			if !ok {
-				return fmt.Errorf("service %q not found", *d.Service.ID)
```
harshadixit12, line 409: @Prashansa-K I wanted to understand if this error is necessary here - while building target state for custom entities during `gateway apply` without lookup tags, the service is not present - so we were returning this.
However, for other entities (say if I'm running `gateway apply` for a route) - we don't throw an error when service is not present.

Prashansa-K: It isn't necessary. `gateway apply` command was built in parallel to this feature. So, these two weren't properly tested together.

Similar checks are also present in `func buildDegraphqlRouteFromCustomEntity`. AFAIR, we fetch "CustomEntities" from dump. So, rawState doesn't have DegraphqlRoutes but CustomEntities are present and processed. How and why does this code block affect the apply command then?

Could you also add some integration tests to account for this? I believe sync tests are already there. We should add apply ones.

harshadixit12: I will add tests. 

> How and why does this code block affect the apply command then?
I made changes to both blocks on local while testing - it's possible I only pushed one. I'll verify.

Also, if rawState doesn't have DegraphqlRoutes, why do we process it [here](https://github.com/Kong/go-database-reconciler/blob/6b9eaf4637969e245871f114149c4fb2d37efb62/pkg/state/builder.go#L402)?
Does it have anything to do with DBless mode support?

harshadixit12:
@Prashansa-K force pushed with two commits - first with only tests, which fails, and second with changes to process raw.Degraphqlroutes (no changes to `buildDegraphqlRouteFromCustomEntity()`) - which passes.
I'll find out why this is happening and open PR for review.

pkg/state/builder.go
Prashansa-K: Adding that check was necessary in `func buildDegraphqlRouteFromCustomEntity` as we don't want to create a degraphqlRoute without a service. Since we can't reinforce the creation of a custom entity via schema, it was necessary to error out here if someone tries to create one without a service. Check this test here which checks for required fields too: https://github.com/Kong/go-database-reconciler/blob/main/pkg/file/builder_test.go#L4010
The same logic was copied above while processing raw.DegraphqlRoutes.


> if rawState doesn't have DegraphqlRoutes, why do we process it [here](https://github.com/Kong/go-database-reconciler/blob/6b9eaf4637969e245871f114149c4fb2d37efb62/pkg/state/builder.go#L402)?

At the moment we don't include DegraphqlRoutes directly in rawState because there's no endpoint that can allow us to fetch them directly. I added it for the sake of completeness. If in future, we have an endpoint that can fetch these routes, we only would have to change the pkg/dump, nothing else.

tests/integration/apply_test.go
```diff
@@ -0,0 +1,43 @@
...
+		t.Fatalf(err.Error())
+	}
+	ctx := context.Background()
+	tests := []struct {
+		name                   string
+		initialStateFile       string
+		targetPartialStateFile string
+	}{
+		{
+			name:                   "certificate",
```
Prashansa-K, line 27: name needs to be corrected here.

harshadixit12:
Pending: PR on deck, and tests on konnect

harshadixit12:
Tested on konnect as well - raised a PR here - https://github.com/Kong/deck/pull/1625

pkg/state/builder.go
Prashansa-K: Yeah. I was giving a reference.
The file package related check would arise only when deck is used. If someone tries directly insert something into state (like we do for intermediate state), then a second check is required. That was my intention of adding it there.

Prashansa-K: So, why do we change it here and not in the `func buildDegraphqlRouteFromCustomEntity`?

harshadixit12: When we are creating `rawState` from `file` package while building target state, we dont store Degraphql routes in CustomEntities field - https://github.com/Kong/go-database-reconciler/blob/main/pkg/file/builder.go#L1892
Which is why the change was made here.

We use `buildDegraphqlRouteFromCustomEntity` when we use `dump.Get` to load current state - in which case degraphql routes are stored in CustomEntities field.

Prashansa-K:
Changes look fine to me. I can approve once the conflict shown is resolved.

Prashansa-K (approved):

PR #271, opened 2025-05-08
fix: fetching plugin-relations from intermediate state too.

Prashansa-K:
### Summary

At the moment, we are fetching
plugin-relations only from
currentState. In scenarios where
the currentState consists of a
global plugin X, syncing
a different state with plugin X
associated with service S
leads to a failure -
"entity already exists".
However, both of these
plugin instances are unique
and should not result in
duplication errors.

This happened because we
were skipping the check
for intermediate state.
The new service created with
second sync would be a part
of intermediate state, not
current state. So, if
we check only currentState, we
miss the serviceID and thus
ingesting a global plugin again,
which led to the duplication
error.

This affected sync, diff and apply commands.

Check https://konghq.atlassian.net/browse/FTI-6648 for reproduction steps.

### Issues resolved

For https://github.com/Kong/deck/issues/1614

### Documentation

### Testing

- [ ] Unit tests
- [X] E2E tests


tests/integration/sync_test.go
```diff
@@ -8052,3 +8052,154 @@ func Test_Sync_Plugins_Nested_Foreign_Keys_EE_3x(t *testing.T) {
...
+func Test_Sync_Scoped_Plugins_Check_Conflicts(t *testing.T) {
+	runWhenKongOrKonnect(t, ">=2.8.0 <3.0.0")
+	setup(t)
+
+	client, err := getTestClient()
+	require.NoError(t, err)
+	ctx := t.Context()
+
+	stateFiles := []string{
+		"testdata/sync/036-scoped-plugins-validation/3x/plugins1.yaml",
```
harshadixit12, line 8144: I see the files we are using are same for tests scoped `>=2.8.0 <3.0.0` and `>=3.0.0` - why do we need different tests?

harshadixit12: The v1 files are unused?
tests/integration/testdata/sync/036-scoped-plugins-validation/plugins1.yaml

harshadixit12 (approved):

PR #274, opened 2025-05-12
feat: added dump and diff logic for keys and keysets

Prashansa-K:
### Issues resolved

For https://github.com/Kong/deck/issues/1444

### Documentation

### Testing

- [ ] Unit tests
- [ ] E2E tests


pkg/dump/dump.go
```diff
@@ -631,6 +661,58 @@ func Get(ctx context.Context, client *kong.Client, config Config) (*utils.KongRa
...
 
+// GetAllKeys queries Kong for all the Keys using client.
+func GetAllKeys(
+	ctx context.Context, client *kong.Client, tags []string,
+) ([]*kong.Key, error) {
+	var keys []*kong.Key
+	opt := newOpt(tags)
+	for {
+		s, nextopt, err := client.Keys.List(ctx, opt)
+		if kong.IsNotFoundErr(err) || kong.IsForbiddenErr(err) {
```
harshadixit12, line 672: Why do we check for not found and forbidden errors for some entities?

harshadixit12 (approved):

pkg/dump/dump.go
Prashansa-K: Some entities may not be a part of all gateway versions. For example, partials was added in `3.10`. So, the List endpoint would give a 404 in earlier versions.
Forbidden error can come up if an entity is meant only for enterprise gateway versions and customer lacks a valid license.
Going forward, we are adding the check for all new entities.

As deck can be used for all supported Kong versions, we have to ensure that false/invalid errors aren't shown to user running a gateway version where some _entity e_ isn't supported or is only meant for paid version.

PR #276, opened 2025-05-13
chore: bump go version to 1.24.3

harshadixit12:
### Summary

Updating go version to 1.24.3
Go v1.24 adds more strict checks for formatted strings - https://tip.golang.org/doc/go1.24

We are replacing instances of `t.Fatalf` and `t.Errorf` used without additional arguments.

### Testing

- [ ] Unit tests
- [x] E2E tests

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,8 +1,6 @@
 module github.com/kong/go-database-reconciler
 
-go 1.23.5
-
-toolchain go1.24.0
+go 1.24.3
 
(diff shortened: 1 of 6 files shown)


go.mod
```diff
@@ -1,8 +1,6 @@
 module github.com/kong/go-database-reconciler
 
-go 1.23.5
-
-toolchain go1.24.0
```
harshadixit12, line 5: toolchain directive is only valid if version > go version

Prashansa-K (approved):

PR #285, opened 2025-05-20
feat: added schema caching for plugins and partials

Prashansa-K:
### Summary

Adding caches for schemas to reduce the number of API calls
to Kong Gateway or Konnect. This would, in turn, speed up
deck operations.

Tested the changes in number of GET calls using the same test file for both the below scenarios.

#### Current deck `v1.47.1`
- No. of GET calls: 415
- No. of GET /schema calls: 393
- No. of GET /schema/plugin calls: 388

#### With caching 
- No. of GET calls: 38
- No. of GET /schema calls: 13
- No. of GET /schema/plugin calls: 8

### Documentation

### Testing

- [ ] Unit tests
- [ ] E2E tests


pkg/types/plugin.go
```diff
@@ -218,6 +222,23 @@ func (d *pluginDiffer) createUpdatePlugin(plugin *state.Plugin) (*crud.Event, er
...
+		return schema, nil
+	}
+
+	schema, err := d.kongClient.Plugins.GetFullSchema(ctx, &pluginName)
+	if err != nil {
+		return schema, err
+	}
+	d.schemasCache[pluginName] = schema
+	return schema, nil
+}
```
pmalek, line 240: This looks awfully similar to `Syncer`'s `getPluginSchema`. I think we could write a separate struct which implements the plugins retrieval and cache and reuse that. WDYT?

Prashansa-K: Done.

pkg/diff/diff.go
```diff
@@ -152,6 +152,12 @@ type Syncer struct {
 
 	// Prevents the Syncer from performing any Delete operations. Default is false (will delete).
 	noDeletes bool
+
+	// schema caching helps in reducing the number of GET calls to Kong Gateway
```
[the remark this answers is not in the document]
Prashansa-K: - Yes. A new Syncer instance is initialised for every new deck command. So, the cache shouldn't grow up to be so large so as to cause an issue. Also, we are caching based on plugin name or partial type. So, repetitions aren't expected.
- Done.

pkg/types/pluginSchemaCache.go
```diff
@@ -0,0 +1,40 @@
...
+
+import (
+	"context"
+	"sync"
+
+	"github.com/kong/go-kong/kong"
+)
+
+type PluginSchemaCache struct {
+	KongClient *kong.Client
```
pmalek, line 11: Ditto

pmalek (approved):
Looks good 👍 . I've checked and our tests run with `-race` so if there's an issue, they should catch it.

PR #288, opened 2025-06-02
fix: false diffs due to ca_cert order in service config

Prashansa-K:
### Summary

A deck sync against Konnect may show a
diff sometimes in a service due to the
ordering difference in deck file and how it
is represented in Konnect. The diff output shown
by deck has no +- signs thus confusing the users.

### Issues resolved

For https://github.com/Kong/deck/issues/1638

### Documentation

### Testing

- [x] Unit tests
- [x] E2E tests


tests/integration/diff_test.go
```diff
@@ -461,7 +461,7 @@ Summary:
 }
 
 `
-	expectedOutputPluginUpdateNoChange = `Summary:
+	expectedOutputNoChange = `Summary:
```
Prashansa-K, line 464: Changed variable name here so that we could reuse it. This wasn't specific to plugins anyway.

PR #289, opened 2025-06-02
fix: fetch partials in case of nested plugins

Prashansa-K:
### Summary

Partial linking for nested plugins was not
working well as it was unaccounted for.
Now, `ingestPlugin()` method links the
partials to the plugins. As this method
is used by other entities that may have a
nested plugin, it would be ensured that all
entities with a nested plugin that use a
partial are able to link it to one.

### Issues resolved

For https://github.com/Kong/deck/issues/1649

### Documentation

### Testing

- [ ] Unit tests
- [ ] E2E tests

### DIFF
--- a/pkg/file/builder.go
+++ b/pkg/file/builder.go
@@ -1559,10 +1559,6 @@ func (b *stateBuilder) plugins() {
 			p.ConsumerGroup = utils.GetConsumerGroupReference(cg.ConsumerGroup)
 		}
 
-		if p.Partials != nil {
-			p.Partials = b.findLinkedPartials(&p.Plugin)
-		}
-
 		if err := b.validatePlugin(p); err != nil {
 			b.err = err
 			return
@@ -1786,7 +1782,7 @@ func (b *stateBuilder) ingestPlugins(plugins []FPlugin) error {
 			}
 		}
 		if p.Partials != nil {
-			plugin.Partials = b.findLinkedPartials(&p.Plugin)
+			p.Partials = b.findLinkedPartials(&p.Plugin)
 		}
 		if p.Config == nil {
 			p.Config = make(map[string]interface{})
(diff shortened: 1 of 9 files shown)


tests/integration/sync_test.go
```diff
@@ -8127,6 +8127,80 @@ func Test_Sync_Partials_Plugins(t *testing.T) {
...
+			CurrentState: currentState,
+			TargetState:  targetState,
+
+			KongClient: client,
+		})
+		require.NoError(t, err)
+
+		stats, errs, changes := syncer.Solve(ctx, 1, false, true)
+		require.Empty(t, errs, "Should have no errors in syncing")
+		logEntityChanges(t, stats, changes)
```
harshadixit12, line 8158: Are we making use of the stats anywhere? 👀

Prashansa-K: We need to log the changes so that test debugging is easier if something blows up.

pkg/file/builder.go
```diff
@@ -1927,6 +1893,44 @@ func (b *stateBuilder) pluginRelations(plugin *kong.Plugin) (cID, rID, sID, cgID
 	return cID, rID, sID, cgID
 }
 
+func (b *stateBuilder) findLinkedPartials(plugin *kong.Plugin) []*kong.PartialLink {
+	var pluginPartials []*kong.PartialLink
+	for _, partial := range plugin.Partials {
+		if partial.Partial == nil {
+			b.err = fmt.Errorf("partial for plugin %v: missing required fields - name or id", *plugin.Name)
```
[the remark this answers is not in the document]
Prashansa-K: I have changed the other error to denote that either of name or id would be needed. Is that better?

harshadixit12 (approved):

PR #296, opened 2025-06-16
fix: added dump config option to skip custom entities if select tags are set

Prashansa-K:
### Summary

Custom entities like degraphql_routes don't
support tagging by the gateway. Thus, if the
gateway is configured with degraphql_routes
and the user tries to sync other configuration
that has "select_tags", failures can come up
if the service referenced in the degraphql_route
isn't a part of the selected configuration.
A sort of workaround is to use default_lookup_tags
to include the referenced service in the state, but it
can lead to the deletion of the degraphql_route which
is undesirable.

Here, we are adding a new dumpConfig option
which lets us skip custom-entities in case select
tags are present. Thus, they don't become a part of
the state and no failures or undesirable deletions occur.

We are keeping this option as an opt-in as this
module is used by both deck and KIC.
Deck would set this option as true.
Ref PR: https://github.com/Kong/deck/pull/1662

### Issues resolved

For https://github.com/Kong/deck/issues/1564

### Documentation

### Testing

- [ ] Unit tests
- [ ] E2E tests


pkg/dump/dump.go
```diff
@@ -588,7 +589,16 @@ func getProxyConfiguration(ctx context.Context, group *errgroup.Group,
 		})
 	}
 
-	if len(config.CustomEntityTypes) > 0 {
+	// If SkipCustomEntitiesWithSelectorTags is true and SelectorTags is not empty,
```
harshadixit12, line 592: If and when custom entities support selector tags, we'd need to revert this, correct?

harshadixit12 (approved):

PR #301, opened 2025-06-24
fix: keys, keysets sync without names

Prashansa-K:
### Summary

keys and key_sets syncing was failing when
the name field was missing from config.
The Gateway allows creation of these entities
without a name.
Thus, this change ensures that name can be
skipped, if required, during creation of keys
and keysets. Name can be added later if
necessary.
Tests are added for both the cases.

Issue discovery: https://kongstrong.slack.com/archives/C03CTMSHP6C/p1750335785138549?thread_ts=1750317075.144389&cid=C03CTMSHP6C

### Issues resolved

For #300 

### Documentation

### Testing

- [ ] Unit tests
- [ ] E2E tests

### DIFF
--- a/tests/integration/sync_test.go
+++ b/tests/integration/sync_test.go
@@ -8775,6 +8775,9 @@ func Test_Sync_Scoped_Plugins_Check_Conflicts(t *testing.T) {
 	}
 }
 
+// test scope:
+//
+// - >=3.1.0
 func Test_Sync_KeysAndKeySets(t *testing.T) {
 	runWhenKongOrKonnect(t, ">=3.1.0")
 	setup(t)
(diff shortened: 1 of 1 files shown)


tests/integration/sync_test.go
```diff
@@ -8787,6 +8787,8 @@ func Test_Sync_KeysAndKeySets(t *testing.T) {
 		name          string
 		kongFile      string
 		expectedState utils.KongRawState
+		update        bool
```
harshadixit12, line 8790: I'm wondering if separating the update into a different test would be a good idea, what do you think @Prashansa-K ?

Prashansa-K: Done.

PR #307, opened 2025-07-02
fix: deck marks false diff due to empty or unordered tags

Prashansa-K:
### Summary
`reflect.DeepEqual()` function returns false
in cases when tags are empty or not in the
same order. Thus, deck marks this as an
update and shows a false diff.
Thus, adding the tag handling code in all
places where it seems missing.

### Issues resolved

For https://github.com/Kong/deck/issues/1674

### Documentation

### Testing

- [ ] Unit tests
- [ ] E2E tests


harshadixit12:
code LGTM, can we add a test for this?

Prashansa-K:
I will add the test in deck. There is too much to and fro right now for testing and this is a minor use-case. So, didn't feel like adding a test here.

PR #303, opened 2025-06-27
feat: add `control_plane_id` field to the dumped configuration

asartori86:
### Summary

Add the field `control_plane_id` to the Konnect schema to support its usage via `deck`

### Full changelog

* Add the field `control_plane_id` to the Konnect schema to support its usage via `deck`

### Issues resolved

n/a

### Documentation

[dependent PR](https://github.com/Kong/deck/pull/1679)

### Testing

- [ ] Unit tests
- [ ] E2E tests
- [ ] Manual testing on Universal
- [ ] Manual testing on Kubernetes

PR #308, opened 2025-07-04
feat: add `control_plane_id` field to the dumped configuration

asartori86:
### Summary

Add the field `control_plane_id` to the Konnect schema to support its usage via `deck`

supersedes #303 

### Full changelog

* Add the field `control_plane_id` to the Konnect schema to support its usage via `deck`

### Issues resolved

n/a

### Documentation

[dependent PR](https://github.com/Kong/deck/pull/1679)

### Testing

Once confirmed that the changes are done in the correct places, I will add more tests

- [ ] Unit tests
- [ ] E2E tests

### DIFF
--- a/pkg/dump/dump.go
+++ b/pkg/dump/dump.go
@@ -58,7 +58,8 @@ type Config struct {
 	LookUpSelectorTagsPartials       []string
 
 	// KonnectControlPlane
-	KonnectControlPlane string
+	KonnectControlPlane   string
+	KonnectControlPlaneID string
 
 	// IsConsumerGroupScopedPluginSupported
 	IsConsumerGroupScopedPluginSupported bool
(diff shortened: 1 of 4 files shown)


asartori86:
Thanks @Prashansa-K for the review. I am a bit confused about the handling of both ID and name. AFIU, the name is also used to support additional logic, e.g., in `pkg/file/reader.go:82`
```go
builder.isKonnect = dumpConfig.KonnectControlPlane != ""
```

If the user didn't specify the name but only the ID, we end up comparing the CP with the provided ID with the (probably present) CP named `default`. If we want to enforce the validity of the pair id-name when both are provided, we should refactor the logic behind the flag `konnect-control-plane-name` so that it can be left empty, allowing us to determine if the control plane name is set to  "default" because the user explicitly specified it.

I could implement the above, but I need some guidance as I am pretty new to deck.

Given that `konnect-control-plane-id` is hidden for now and mutually exclusive with the `*-name` flags, the most straightforward approach would be to either dump the control plane name or the ID, not both. If the name is not specified in the config, will any logic break?

PR #352, opened 2025-10-17
feat: skip-defaults in dump

Prashansa-K:
### Summary

We are adding a new feature for `deck gateway dump` where
a user can choose to skip default values from getting dumped.
We use the schemas fetched from the gateway to check for default
values and remove them from the dump config.
This works only when default values are set in schema. Hence, a few
values (like shorthand_fields in redis for some plugins) may still show
up if defaults are missing in schema. We will handle these at a later stage.
Schema caching is added to avoid multiple API calls.

Unit tests for new functions are added. Integration tests would be
added in a separate PR.

### Issues resolved

For https://github.com/Kong/deck/issues/1456

### Documentation

### Testing

- [ ] Unit tests
- [ ] E2E tests

### DIFF
--- a/pkg/dump/skip_defaults.go
+++ b/pkg/dump/skip_defaults.go
@@ -13,8 +13,10 @@ import (
 	"golang.org/x/sync/errgroup"
 )
 
-var defaultFieldsCache = map[string]interface{}{}
-var defaultFieldsMu sync.Mutex
+var (
+	defaultFieldsCache = map[string]interface{}{}
+	defaultFieldsMu    sync.Mutex
+)
 
 func removeDefaultsFromState(ctx context.Context, group *errgroup.Group,
 	state *utils.KongRawState, schemaFetcher *SchemaFetcher,


pkg/dump/skip_defaults_test.go
```diff
@@ -0,0 +1,489 @@
...
+					"timeout":   6000,
+					"protocols": []string{"grpc", "https"},
+				},
+			},
+		},
+		{
+			name: "field has nil value",
+			fieldMap: map[string]interface{}{
+				"timeout": 5000,
+				"name":    nil,
```
harshadixit12: If `name` has a non-null default - and user has configured it to be null - does it still get removed?

pkg/dump/schemas.go
```diff
@@ -0,0 +1,165 @@
...
+	"context"
+	"fmt"
+	"net/http"
+
+	"github.com/kong/go-database-reconciler/pkg/types"
+	"github.com/kong/go-kong/kong"
+)
+
+// Some entities in Konnect have different names compared to Kong Gateway
+var kongToKonnectEntitiesMap = map[string]string{
```
harshadixit12: /pkg/utils/defaulter.go has something similar - can we reuse?

pkg/dump/skip_defaults.go
```diff
@@ -0,0 +1,578 @@
...
+			defaultFields[fname] = value.Value()
+		}
+
+		return true
+	})
+
+	return defaultFields
+}
+
+func parseEntityWithDefaults(entity reflect.Value, defaultFields map[string]interface{}) error {
```
harshadixit12: nitpick - function name could indicate we are removing defaults wherever it makes sense?

harshadixit12 (approved):

PR #358, opened 2025-11-17
Fix/no diff request transformer plugin

harshadixit12:
### Summary
In this PR, we are modifying the nestedArraySort function to accept plugin schema, and sort based on the `type`.
For type `array`, we do not sort, but for type `set` we sort.
The implementation is designed to fall back to original behaviour (sort) when schema is unavailable.

### Risks & mitigation
1. No schema given for a field - in this case, the behaviour remains as is - sort.
2. Performance - the new implementation for comparing plugin equality is slower - but since the function only takes few microseconds per plugin, it adds negligible degradation overall.
3. Diff shown - the diff is missing `+-` in array reorder - due to the [json diffing tool](https://github.com/Kong/gojsondiff) we use. I have called out to check if it is okay here - [link](https://kongstrong.slack.com/archives/C07AQH7SAF8/p1763534009205859)

### Full changelog

* Fix https://github.com/Kong/deck/issues/1741

### Issues resolved

Fix https://github.com/Kong/deck/issues/1741

### Documentation

### Testing

- [x] Unit tests
- [x] E2E tests

Deck PR with passing tests: https://github.com/Kong/deck/pull/1820 (original - https://github.com/Kong/deck/pull/1819, but has a failing test on konnect due to changes in a different plugin schema)

### DIFF
--- a/pkg/state/types.go
+++ b/pkg/state/types.go
@@ -12,12 +12,6 @@ import (
 	"github.com/tidwall/gjson"
 )
 
-const (
-	pluginConfigKey              = "fields.#(config).config"
-	fieldsQueryTemplate          = "fields.#(%s).%s"
-	shorthandFieldsQueryTemplate = "shorthand_fields.#(%s).%s"
-)
-
 // entity abstracts out common fields in a credentials.
 // TODO generalize for each and every entity.
 type entity interface {
@@ -597,6 +591,7 @@ func (p1 *Plugin) EqualWithOpts(p2 *Plugin, ignoreID,
 	sort.Slice(p1Copy.Protocols, func(i, j int) bool { return *(p1Copy.Protocols[i]) < *(p1Copy.Protocols[j]) })
 	sort.Slice(p2Copy.Protocols, func(i, j int) bool { return *(p2Copy.Protocols[i]) < *(p2Copy.Protocols[j]) })
 
+	const pluginConfigKey = "fields.#(config).config"
 	configSchema := schema.Get(pluginConfigKey)
 	p1Copy.Config = sortNestedArraysBasedOnSchema(p1Copy.Config, configSchema)
 	p2Copy.Config = sortNestedArraysBasedOnSchema(p2Copy.Config, configSchema)
@@ -682,6 +677,8 @@ func (e EmptyInterfaceUsingUnderlyingType) Less(i, j int) bool {
 
 // Helper function to get schema for a field name
 func getSchemaForFieldName(schema gjson.Result, fieldName string) gjson.Result {
+	const fieldsQueryTemplate = "fields.#(%s).%s"
+	const shorthandFieldsQueryTemplate = "shorthand_fields.#(%s).%s"
 	fieldsQuery := fmt.Sprintf(fieldsQueryTemplate, fieldName, fieldName)
 	result := schema.Get(fieldsQuery)
 	if !result.Exists() {


pkg/state/types.go
```diff
@@ -565,14 +566,14 @@ func (p1 *Plugin) Console() string {
 // Equal returns true if r1 and r2 are equal.
 // TODO add compare array without position
 func (p1 *Plugin) Equal(p2 *Plugin) bool {
```
harshadixit12: For plugin, this function didn't seem to be used directly anywhere.
However, it makes sense to expect schema as org here as well - will make that change in next pass.

pkg/state/types.go
```diff
@@ -672,26 +673,59 @@ func (e EmptyInterfaceUsingUnderlyingType) Less(i, j int) bool {
...
+	typeResult := schema.Get("type")
+	if !typeResult.Exists() {
+		return false
+	}
+
+	return typeResult.String() == "array"
+}
+
+// Helper function to sort nested arrays in a map referring to schema
+func sortNestedArraysBasedOnSchema(m map[string]interface{}, schema gjson.Result) map[string]interface{} {
```
harshadixit12: This implementation of sort is slower than previous, by 10x based on benchmark. However, it still takes few microseconds, which will be negligible compared to rest.

pkg/types/plugin.go
```diff
@@ -209,7 +211,13 @@ func (d *pluginDiffer) createUpdatePlugin(plugin *state.Plugin) (*crud.Event, er
 		return nil, fmt.Errorf("failed clearing unmatching deprecations fields: %w", err)
 	}
 
-	if !currentPlugin.EqualWithOpts(pluginWithDefaults, false, true, false) {
+	jsonb, err := json.Marshal(&schema)
+	if err != nil {
```
Prashansa-K: In case of an error, should we consider proceeding to comparison with nil schema object? Error here doesn't seem significant enough to stop the entire process.

pkg/types/plugin.go
```diff
@@ -209,7 +211,13 @@ func (d *pluginDiffer) createUpdatePlugin(plugin *state.Plugin) (*crud.Event, er
...
+		return nil, err
+	}
+	gjsonSchema := gjson.ParseBytes((jsonb))
```
Prashansa-K: Curious: Why double brackets?

pkg/state/types.go
Prashansa-K: As far as naming convention goes, any "options" should be passed as args in EqualWithOpts function only. So, we should keep this one as it is.

pkg/state/types.go
```diff
@@ -590,8 +591,8 @@ func (p1 *Plugin) EqualWithOpts(p2 *Plugin, ignoreID,
 	sort.Slice(p1Copy.Protocols, func(i, j int) bool { return *(p1Copy.Protocols[i]) < *(p1Copy.Protocols[j]) })
 	sort.Slice(p2Copy.Protocols, func(i, j int) bool { return *(p2Copy.Protocols[i]) < *(p2Copy.Protocols[j]) })
 
-	p1Copy.Config = sortNestedArrays(p1Copy.Config)
-	p2Copy.Config = sortNestedArrays(p2Copy.Config)
+	p1Copy.Config = sortNestedArraysBasedOnSchema(p1Copy.Config, schema.Get("fields.#(config).config"))
```
Prashansa-K: Let's assign a configSchema once and use it in both instances.

pkg/state/types.go
```diff
@@ -672,26 +673,59 @@ func (e EmptyInterfaceUsingUnderlyingType) Less(i, j int) bool {
 	return strings.Compare(objI, objJ) == -1
 }
 
-// Helper function to sort nested arrays in a map
-func sortNestedArrays(m map[string]interface{}) map[string]interface{} {
-	sortedMap := make(map[string]interface{})
+// Helper function to get schema for a field name
+func getSchemaForFieldName(schema gjson.Result, fieldName string) gjson.Result {
+	fieldsQuery := "fields.#(" + fieldName + ")." + fieldName
```
Prashansa-K: Let's use constants for the strings that we don't want to be mutated like `"fields.#("`. Makes the code more readable too.

pkg/state/types.go
```diff
@@ -672,26 +673,59 @@ func (e EmptyInterfaceUsingUnderlyingType) Less(i, j int) bool {
...
+	if !result.Exists() {
+		// try shorthand fields
+		shorthandQuery := "shorthand_fields.#(" + fieldName + ")." + fieldName
+		result = schema.Get(shorthandQuery)
+	}
+	return result
+}
+
+// Helper function to determine if we should skip sorting based on schema
+func skipSort(schema gjson.Result) bool {
```
Prashansa-K: Suggestion:
Instead of skipSort that negates the outcome of this function during usage, what if we use `shouldSort` instead? Makes the code flow more natural. WDYT?

harshadixit12: Makes sense 👍

pkg/state/types.go
```diff
@@ -9,6 +9,13 @@ import (
 	"strings"
 
 	"github.com/kong/go-kong/kong"
+	"github.com/tidwall/gjson"
+)
+
+const (
```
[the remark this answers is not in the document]
Prashansa-K: Let's make them local only for now. These are too specific to be required elsewhere.

harshadixit12:
> Is this is a single run comparison or a benchmark?

With parallelism=1, it was 3 manual runs of diff command, using `time` in zsh - gave same result. This was done to measure performance on a large real world config.

PR #366, opened 2026-01-23
fix: instance_name field added to ConsumerGroupPlugin

shivaygupta-dotcom:
This pull request adds support for the `instance_name` field to consumer group-scoped plugins throughout the codebase. It ensures that plugins attached to consumer groups can specify an `instance_name`, updates the schema, builder logic, and tests, and verifies the feature with new integration tests and test data.

**Consumer group plugin enhancements:**

* Added support for the `instance_name` field in consumer group-scoped plugins in both the builder logic (`pkg/file/builder.go`) and writer logic (`pkg/file/writer.go`). This allows plugins attached to consumer groups to specify an `instance_name`. 
* Updated the JSON schema for consumer group plugins to include the `instance_name` property, ensuring proper validation and documentation.

**Testing and validation:**

* Added and updated unit tests in `pkg/file/builder_test.go` to cover scenarios where consumer group plugins have an `instance_name` set, improving test coverage and reliability.
* Added integration tests and test data to verify syncing and dumping of consumer group plugins with `instance_name`, ensuring end-to-end correctness. 
**Test data updates:**

* Updated multiple test data files to include the `instance_name` field for consumer group plugins, reflecting the new feature in expected outputs. 

**Dependency update:**

* Updated the `go-kong` dependency to a newer commit to support the `instance_name` feature for consumer group plugins.


pkg/file/builder_test.go
```diff
@@ -5597,3 +5599,99 @@ func Test_stateBuilder_ingestConsumerGroupConsumer(t *testing.T) {
...
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(_ *testing.T) {
+			ctx := context.Background()
+			b := &stateBuilder{
+				targetContent: tt.fields.targetContent,
+				currentState:  tt.fields.currentState,
+				kongVersion:   kong340Version,
```
Prashansa-K: Can we check this for multiple kong versions? A nested loop inside the range tests loops which checks this should be fine. Let's check for 3.4, 3.7, 3.8, 3.9, 3.10, 3.11 for now

pkg/file/builder_test.go
```diff
@@ -4443,7 +4443,8 @@ func Test_stateBuilder_ConsumerGroupPolicyOverrides(t *testing.T) {
 							Consumers: nil,
 							Plugins: []*kong.ConsumerGroupPlugin{
 								{
-									Name: kong.String("rate-limiting-advanced"),
+									Name:         kong.String("rate-limiting-advanced"),
+									InstanceName: kong.String("custom-instance-name"),
```
Prashansa-K: The build function should ideally fail here. 🤔 
If `ConsumerGroupPolicyOverrides` is set, we can't set an instance_name. 

For now, could you remove these changes from this test function and file a bug in this repo regarding the same? We can take it up in the next iteration.

tests/integration/testdata/apply/006-consumer-group-plugins/consumer-group-plugin-initial.yaml
```diff
@@ -0,0 +1,14 @@
...
+    plugins:
+      - config:
+          namespace: gold
+          limit:
+            - 7
+          retry_after_jitter_max: 1
+          window_size:
+            - 60
+          window_type: sliding
+        instance_name: default-instance
```
Prashansa-K: If both initial and final files have the instance name set, then what exactly are we checking?
The purpose of the apply test here is to check that instance_name can be added later on by users if required.

tests/integration/apply_test.go
```diff
@@ -373,3 +373,106 @@ func Test_Apply_Consumer_Group_Consumer(t *testing.T) {
 		})
 	}
 }
+
+func Test_Apply_Consumer_Group_Plugin(t *testing.T) {
+	runWhen(t, "enterprise", ">=3.10.0")
```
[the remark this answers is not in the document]
shivaygupta-dotcom: done.

PR #428, opened 2026-03-11
feat: Added changes to support crud for graphql-cost-decoration

shivaygupta-dotcom:
https://konghq.atlassian.net/browse/FTI-7307
https://github.com/Kong/deck/issues/1909

### Summary
This PR adds full CRUD support for the graphql_ratelimiting_cost_decorations custom entity used by Kong's GraphQL Rate Limiting Advanced plugin.

### Full changelog

1. Added state management for graphql_ratelimiting_cost_decorations by adding support for collecting entities, indexing them by ID and TypePath, and enabling CRUD operations in the state package so decorations can be properly tracked during sync and diff operations.
2. Added the new entity into the sync/diff pipeline by registering it in the differ, post-processor, and entity order definitions, ensuring decorations are correctly created, updated, and deleted during deck sync and deck apply.
3. Added dump support using the CustomEntityTypes configuration, along with custom handling to map the entity type to Kong’s actual API endpoint (/graphql-rate-limiting-advanced/costs).
4. Added integration tests covering sync, apply, and dump operations, including scenarios for basic usage, multiple decorations, full field coverage, and error cases.

### Issues resolved

Fix #XXX

### Documentation

### Testing

- [ ] Unit tests
- [ ] E2E tests


pkg/file/builder.go
```diff
@@ -2291,6 +2294,147 @@ func (b *stateBuilder) copyToDegraphqlRoute(fcEntity FCustomEntity) (DegraphqlRo
 	return degraphqlRoute, nil
 }
 
+func (b *stateBuilder) ingestGraphqlRateLimitingCostDecoration(entity FCustomEntity) {
```
Prashansa-K, line 2331: We will require unit tests for the changes in this file in file/builder_tests.go

pkg/utils/tags_test.go
```diff
@@ -49,7 +49,13 @@ func equalArray(want, have []*string) bool {
 	if len(want) != len(have) {
 		return false
 	}
-	for i := 0; i < len(want); i++ {
+	for i := range want {
+		if want[i] == nil || have[i] == nil {
```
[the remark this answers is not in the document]
Prashansa-K: Do we still need this here? The mentioned PR got merged.

pkg/file/types.go
```diff
@@ -1006,6 +1020,19 @@ type DegraphqlRoute struct {
 	kong.DegraphqlRoute
 }
 
+// +k8s:deepcopy-gen=true
```
Prashansa-K, line 1023: Run the generator so that we can see changes in the auto-gen file.

pkg/file/types.go
```diff
@@ -973,6 +974,13 @@ func (f *FCustomEntity) UnmarshalJSON(b []byte) error {
 			return err
 		}
 		return copyToFCustomEntity(entity, f)
+	case graphqlRateLimitingCostDecorationsType:
+		var entity map[string]interface{}
+		err := json.Unmarshal(b, &entity)
+		if err != nil {
+			return err
+		}
+		return copyToGraphqlRateLimitingCostDecorationEntity(entity, f)
```
Prashansa-K: Can you make the func names shorter so that they read better?

`copyToGqlRateLimitingCostDecoration` and `copyFromGqlRateLimitingCostDecoration`

pkg/state/builder.go
```diff
@@ -585,3 +606,87 @@ func buildDegraphqlRouteFromCustomEntity(kongState *KongState, entity map[string
...
+	if entity["id"] != nil {
+		id, ok := entity["id"].(string)
+		if !ok {
+			return GraphqlRateLimitingCostDecoration{}, fmt.Errorf("id must be of type string")
+		}
+		decoration.ID = kong.String(id)
+	}
+
+	if entity["service"] != nil {
+		svc, ok := entity["service"].(map[string]interface{})
```
Prashansa-K, line 632: What if someone passes `service: svc1` => service passed as a string?

Prashansa-K: This should be caught and an apt error should be thrown

pkg/state/builder.go
```diff
@@ -585,3 +606,87 @@ func buildDegraphqlRouteFromCustomEntity(kongState *KongState, entity map[string
...
+		if ok {
+			decoration.Service = &kong.Service{}
+			if id, ok := svc["id"].(string); ok {
+				decoration.Service.ID = kong.String(id)
+			}
+			if name, ok := svc["name"].(string); ok {
+				decoration.Service.Name = kong.String(name)
+			}
+		}
```
Prashansa-K, line 654: We also need to do a ensureService() check to make sure that the passed service exists.

pkg/state/builder.go
```diff
@@ -585,3 +606,87 @@ func buildDegraphqlRouteFromCustomEntity(kongState *KongState, entity map[string
...
+	}
+
+	if entity["service"] != nil {
+		svc, ok := entity["service"].(map[string]interface{})
+		if ok {
+			decoration.Service = &kong.Service{}
+			if id, ok := svc["id"].(string); ok {
+				decoration.Service.ID = kong.String(id)
+			}
+			if name, ok := svc["name"].(string); ok {
```
Prashansa-K: At this point, would you ever get "name" for service? 🤔 
You are pushing in ID only from file builder: https://github.com/Kong/go-database-reconciler/pull/428/changes#diff-c176c7e7e3271c45169d0fd8cb7f2adb1c06292d2ce705710a304f8bc4acf4feR2369

pkg/utils/tags_test.go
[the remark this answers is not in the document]
shivaygupta-dotcom: Some git sync issue, I can see these changes are already merged in main branch.

pkg/state/builder.go
shivaygupta-dotcom: yes you are right, name will not be present as of current implementation, removed it.

pkg/state/builder.go
```diff
@@ -585,3 +606,100 @@ func buildDegraphqlRouteFromCustomEntity(kongState *KongState, entity map[string
...
+	if entity["service"] != nil {
+		svc, ok := entity["service"].(map[string]interface{})
+		if !ok {
+			return GraphqlRateLimitingCostDecoration{}, fmt.Errorf("service must be of type object")
+		}
+
+		serviceID, ok := svc["id"].(string)
+		if !ok {
+			return GraphqlRateLimitingCostDecoration{},
+				fmt.Errorf("service must be of type object with a valid string id")
```
Prashansa-K: "service must be of type object with a valid string id or name."
Let's keep this similar to degraphql routes

pkg/state/builder.go
```diff
@@ -585,3 +606,100 @@ func buildDegraphqlRouteFromCustomEntity(kongState *KongState, entity map[string
...
+		}
+
+		ok, s, err := ensureService(kongState, serviceID)
+		if err != nil {
+			return GraphqlRateLimitingCostDecoration{}, err
+		}
+		if !ok {
+			return GraphqlRateLimitingCostDecoration{},
+				fmt.Errorf("service must be of type object with a valid string id")
```
Prashansa-K: Same as L631

pkg/state/builder.go
```diff
@@ -431,8 +431,17 @@ func buildKong(kongState *KongState, raw *utils.KongRawState) error {
 		}
 	}
 
+	for _, g := range raw.GraphqlRateLimitingCostDecorations {
+		err := kongState.GraphqlRateLimitingCostDecorations.Add(
```
Prashansa-K, line 444: You will need to check for service here too, right?
Service existence + validity via ensureService()? We are checking for the same while using "custom entities".

Can a graphql cost decorator exist without a service?

pkg/state/graphql_ratelimiting_cost_decoration_test.go
```diff
@@ -0,0 +1,282 @@
...
+	err := collection.Add(decoration)
+	require.NoError(err, "error adding decoration")
+
+	// Fetch the currently added entity
+	res, err := collection.Get("example")
+	require.NoError(err, "error getting decoration")
+	require.NotNil(res)
+	assert.Equal("example", *res.ID)
+	assert.Equal("Query.users", *res.TypePath)
+	assert.InDelta(1.0, *res.AddConstant, 0.001)
```
Prashansa-K: Why inDelta with 0.001 margin? 
If we are adding "1.0", we should expect "1.0" exactly right?

Prashansa-K: Any delta would make deck show a diff.

Prashansa-K: Same question for other instances of inDelta in other test functions.

pkg/state/types.go
```diff
@@ -1972,6 +1972,55 @@ func (d *DegraphqlRoute) EqualWithOpts(d2 *DegraphqlRoute, ignoreID bool) bool {
...
+}
+
+// GetCustomEntityType returns the GraphqlRateLimitingCostDecoration Type.
+func (g *GraphqlRateLimitingCostDecoration) GetCustomEntityType() string {
+	return "graphql_ratelimiting_cost_decorations"
+}
+
+// Console returns the string to identify the GraphqlRateLimitingCostDecoration.
+func (g *GraphqlRateLimitingCostDecoration) Console() string {
+	if g.TypePath != nil {
```
Prashansa-K: Why would we print TypePath during a Console() call?
It's usually name or ID. Is TypePath equivalent to a name field?

pkg/types/graphql_ratelimiting_cost_decoration.go
```diff
@@ -0,0 +1,204 @@
...
+
+// Create creates a GraphqlRateLimitingCostDecoration in Kong.
+// The arg should be of type crud.Event, containing the decoration to be created,
+// else the function will panic.
+// It returns the created *state.GraphqlRateLimitingCostDecoration.
+func (s *graphqlRateLimitingCostDecorationCRUD) Create(ctx context.Context, arg ...crud.Arg) (crud.Arg, error) {
+	event := crud.EventFromArg(arg[0])
+	decoration := graphqlRateLimitingCostDecorationFromStruct(event)
+
+	createdDecoration, err := s.client.GraphqlRateLimitingCostDecorations.CreateForService(ctx,
```
Prashansa-K, line 43: Check if stripping service name is necessary or not, specially for Konnect.
Konnect doesn't allow passing names for foreign keys. In that case, we need to just keep the IDs.
Check how we do this in degraphql routes.

pkg/types/graphql_ratelimiting_cost_decoration.go
```diff
@@ -0,0 +1,204 @@
...
+	if err == nil {
+		// Found by ID, no delete needed
+		return nil, nil
+	}
+	if !errors.Is(err, state.ErrNotFound) {
+		return nil, fmt.Errorf("looking up graphql ratelimiting cost decoration %q: %w", *decoration.ID, err)
+	}
+
+	// Not found by ID, try to find by TypePath
+	if decoration.TypePath != nil {
```
Prashansa-K: Why?
Nowhere in other entities can I see a "find by other unique key". We use IDs only. Why are we doing it differently here?

pkg/types/graphql_ratelimiting_cost_decoration.go
```diff
@@ -0,0 +1,204 @@
...
+
+	// First try to find by ID
+	currentDecoration, err := d.currentState.GraphqlRateLimitingCostDecorations.Get(*decoration.ID)
+	if err != nil && !errors.Is(err, state.ErrNotFound) {
+		return nil, fmt.Errorf("error looking up graphql ratelimiting cost decoration %q: %w",
+			*decoration.ID, err)
+	}
+
+	// If not found by ID, try to find by TypePath
+	if errors.Is(err, state.ErrNotFound) && decoration.TypePath != nil {
```
Prashansa-K: Same question as L119

pkg/file/builder.go
```diff
@@ -2291,6 +2294,147 @@ func (b *stateBuilder) copyToDegraphqlRoute(fcEntity FCustomEntity) (DegraphqlRo
...
+		decoration.ID = fcEntity.ID
+	}
+
+	if fcEntity.Fields == nil {
+		return GraphqlRateLimitingCostDecoration{},
+			fmt.Errorf("fields are required for graphql_ratelimiting_cost_decorations")
+	}
+
+	if fcEntity.Fields["id"] != nil {
+		if id, ok := fcEntity.Fields["id"].(*string); ok {
```
Prashansa-K: Why would id be nested inside fields?

What's the format?

```
custom_entities:
-  id: my-id
    type: my-type
    fields:
      service:
          id: my-svc-id
          name: my-svc
      other_field: value
```

pkg/file/builder.go
```diff
@@ -2291,6 +2294,147 @@ func (b *stateBuilder) copyToDegraphqlRoute(fcEntity FCustomEntity) (DegraphqlRo
...
+					addArgs[i] = kong.String(argStr)
+				} else if argPtr, ok := arg.(*string); ok {
+					addArgs[i] = argPtr
+				}
+			}
+			decoration.AddArguments = addArgs
+		}
+	}
+
+	if fcEntity.Fields["mul_arguments"] != nil {
```
Prashansa-K, line 2439: I didn't get this.
What's the type for mul_arguments?
Is it a string array or a stringPtr array? How can it be both at the same time?

pkg/file/builder.go
```diff
@@ -2291,6 +2294,147 @@ func (b *stateBuilder) copyToDegraphqlRoute(fcEntity FCustomEntity) (DegraphqlRo
...
+				} else if argPtr, ok := arg.(*string); ok {
+					mulArgs[i] = argPtr
+				}
+			}
+			decoration.MulArguments = mulArgs
+		}
+	}
+
+	if decoration.TypePath == nil {
+		return GraphqlRateLimitingCostDecoration{},
```
Prashansa-K, line 2450: Move this just after the Fields check to fail early.

pkg/file/builder_test.go
```diff
@@ -5948,3 +5968,261 @@ func Test_InstanceName_ConsumerGroupPlugin(t *testing.T) {
...
+			},
+			want: &utils.KongRawState{
+				GraphqlRateLimitingCostDecorations: nil,
+			},
+			wantErr:   true,
+			errString: "fields are required for graphql_ratelimiting_cost_decorations",
+		},
+	}
+
+	for _, tt := range tests {
```
Prashansa-K, line 6430: Add some test for fields like mul_arguments and add_arguments.

pkg/dump/dump.go
```diff
@@ -644,19 +644,42 @@ func getProxyConfiguration(ctx context.Context, group *errgroup.Group,
...
+		}
+
 		customEntityLock := sync.Mutex{}
 		for _, entityType := range config.CustomEntityTypes {
 			t := entityType
+
+			// In Konnect mode, graphql_ratelimiting_cost_decorations must be fetched
+			// per-service using the typed ListAllForService endpoint, because Konnect
+			// does not expose the generic custom entities list for this type.
+			if config.KonnectControlPlane != "" && t == "graphql_ratelimiting_cost_decorations" {
```
Prashansa-K: This should be a const: "graphql_ratelimiting_cost_decorations"

pkg/dump/dump.go
```diff
@@ -644,19 +644,42 @@ func getProxyConfiguration(ctx context.Context, group *errgroup.Group,
...
+				group.Go(func() error {
+					decorations, err := GetAllGraphqlRateLimitingCostDecorationsForServices(
+						ctx, client, config.SelectorTags)
```
Prashansa-K: This entity doesn't support tags. Why would you pass select tags here?
Also, when select tags are passed, we skip the custom entities anyway.

tests/integration/apply_test.go
```diff
@@ -30,12 +30,22 @@ func Test_Apply_Custom_Entities(t *testing.T) {
...
+			name:                   "custom entity - graphql_ratelimiting_cost_decorationss multiple",
+			initialStateFile:       "testdata/apply/001-custom-entities/initial-state.yaml",
+			targetPartialStateFile: "testdata/apply/001-custom-entities/graphql-cost-decoration-multiple.yaml",
+		},
 	}
 
 	for _, tc := range tests {
 		t.Run(tc.name, func(t *testing.T) {
 			mustResetKongState(ctx, t, client, deckDump.Config{})
-			err := sync(tc.initialStateFile)
```
Prashansa-K, line 38: Why did we remove the initialisation here?

tests/integration/dump_test.go
```diff
@@ -416,6 +416,131 @@ func Test_Dump_CustomEntities(t *testing.T) {
...
+	}
+
+	for _, tc := range tests {
+		t.Run(tc.name, func(t *testing.T) {
+			reset(t)
+			require.NoError(t, sync("testdata/sync/001-create-a-service/kong3x.yaml"))
+
+			// Create decorations for this test case
+			for _, deco := range tc.decorations {
+				_, err := client.GraphqlRateLimitingCostDecorations.CreateForService(
```
Prashansa-K: Use "sync" or "apply" commands for creations rather than direct client calls wherever necessary. These tend to fail in tests and don't get retried by default.

tests/integration/dump_test.go
```diff
@@ -416,6 +416,131 @@ func Test_Dump_CustomEntities(t *testing.T) {
 	require.Equal(t, "query{ name }", query)
 }
 
+func Test_Dump_GraphqlRateLimitingCostDecorations(t *testing.T) {
```
Prashansa-K: Use "dump" command and output content to a file too see the command's working here.

The other test you see for degraphql route predates deck support for degraphql. It was used for KO or KIC testing. So, don't follow that testing way.

Focus on deck's testing here.

pkg/state/builder.go
shivaygupta-dotcom: > Can a graphql cost decorator exist without a service?

Actually it can from API perspective. but the way we are handling in decK, we need services in every implementation apart from delete for Konnect.

Adding the fix which you suggested.

pkg/types/graphql_ratelimiting_cost_decoration.go
shivaygupta-dotcom: Actually my bad, I have tested one scenario with API where if id is not present but type_path is present, entity was getting created successfully. 
Just now verified again, entity is getting created with same type path multiple times. 
removing this, thanks for the catch.

pkg/types/graphql_ratelimiting_cost_decoration.go
shivaygupta-dotcom: sorry, I didn't understood this comment, we are anyways just passing id in the service.

pkg/file/builder.go
shivaygupta-dotcom: I have added id field two times by mistakes,  while adding fields for every other fields.
removed `fcEntity.Fields["id"]`check, kept only golbal one.

pkg/types/graphql_ratelimiting_cost_decoration.go
shivaygupta-dotcom: pls ignore prev comment.

pkg/state/types.go
shivaygupta-dotcom: https://github.com/Kong/go-database-reconciler/pull/428#discussion_r3110016757

pkg/types/graphql_ratelimiting_cost_decoration.go
shivaygupta-dotcom: https://github.com/Kong/go-database-reconciler/pull/428#discussion_r3110016757

tests/integration/dump_test.go
shivaygupta-dotcom: sync doesnt support custom entities, we have test written in similar way for degraphql_routes as well.
Test_Dump_CustomEntities

tests/integration/apply_test.go
shivaygupta-dotcom: we already had one err var defined above, so used that var below instead of creating new one.
this happened with IDE formatting.

tests/integration/dump_test.go
shivaygupta-dotcom: dump doesnt support custom entities, we have test written in similar way for degraphql_routes as well.
Test_Dump_CustomEntities.

pkg/file/builder.go
shivaygupta-dotcom: removed handling with interfaces, and kept it same as degraphql_routes handling for similar type of data types.

tests/integration/dump_test.go
Prashansa-K: > dump doesn't support custom entities

??
You have to pass a dumpConfig with supported entities. 

> we have test written in similar way for degraphql_routes as well.

The intent for that was different as I mentioned in my previous comment. You can check that test's git history.

If adding a dump test is difficult here, remove it and add it in deck repo later. Don't rely on client though. It makes the tests flaky.

pkg/state/graphql_ratelimiting_cost_decoration_test.go
```diff
@@ -168,7 +168,7 @@ func TestGraphqlRateLimitingCostDecorationUpdate(t *testing.T) {
 	res, err := collection.Get("update-example")
 	require.NoError(err, "error getting decoration")
 	require.NotNil(res)
-	assert.InDelta(1.0, *res.AddConstant, 0.001)
+	assert.InDelta(1.0, *res.AddConstant, 1.0)
```
Prashansa-K: Why would we need a delta here?

This means *res.AddConstant is in the range [0.0, 2.0].
If I define *res.AddConstant = 1.0 in deck, shouldn't I expect it equal to "1.0" when creating the internal state?

pkg/types/graphql_ratelimiting_cost_decoration.go
Prashansa-K: This is not removed. Still in diff.
Please remove this block.

tests/integration/dump_test.go
shivaygupta-dotcom: Actually, I have tried using dump command but it was not working. read comment on `Test_Dump_CustomEntities` and it was mentioned that custom entities was not supported.
I will remove test from here and keep in deck only.

pkg/state/graphql_ratelimiting_cost_decoration_test.go
shivaygupta-dotcom: for this when I am not using plain equal, I am getting linter error
either I can use //nolint for this line or I can add Indelta/InEpsilon for 0.0001.

is it fine, if I use 0.0001 with Indelta/InEpsilon?

ref: https://github.com/Antonboom/testifylint#float-compare

Prashansa-K (approved):

PR #444, opened 2026-04-07
feat: replace mockEnvVars bool with RenderEnvVarsMode in GetContentFromFiles

mheap:
## Summary

Adds new`GetContentFromFilesWithEnvVars` method with a typed `RenderEnvVarsMode`, adding a new `EnvVarsSkip` mode that skips template rendering entirely.

## Changes

- Added `RenderEnvVarsMode` type with three constants to `pkg/file/template.go`:
  - `EnvVarsExpand` — expands env vars (replaces `false`)
  - `EnvVarsMock` — returns var names as values (replaces `true`)
  - `EnvVarsSkip` — skips template rendering entirely (new)
- Updated `GetContentFromFilesWithEnvVars`, `getContent`, `readContent`, and `renderTemplate` to use the new type
- Updated all callers in tests and integration tests

## Usage

```go
// Skip env var expansion entirely
content, err := file.GetContentFromFilesWithEnvVars(filenames, file.EnvVarsSkip)
```

### DIFF
--- a/pkg/file/reader.go
+++ b/pkg/file/reader.go
@@ -33,14 +33,29 @@ type RenderConfig struct {
 // rooted at filename, read all the files with .yaml, .yml and .json extensions
 // and generate a content after a merge of the content from all the files.
 //
+// If mockEnvVars is true, environment variables are replaced with their names
+// instead of their values while rendering templates.
+//
 // It will return an error if the file representation is invalid
 // or if there is any error during processing.
 func GetContentFromFiles(filenames []string, mockEnvVars bool) (*Content, error) {
+	mode := EnvVarsExpand
+	if mockEnvVars {
+		mode = EnvVarsMock
+	}
+
+	return GetContentFromFilesWithEnvVars(filenames, mode)
+}
+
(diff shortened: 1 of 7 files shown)


Prashansa-K:
> Note: This is a breaking API change — callers must update from bool to RenderEnvVarsMode

We will have to push for major version release in that case, which would require a lot of changes. Is this really worth the effort?
What is the exact usecase for this?

mheap:
@Prashansa-K I don't think we need the major version bump here. There is only one consumer (deck) which we also own.

If you feel strongly, I can add a new method which has the new signature and make the old one an alias

Prashansa-K:
What if we make a new config struct for this or utilise the existing `RenderConfig` struct to add the new modes? deck can set the config and the GetContent* interfaces can use them without breaking contracts?

We use a similar model for dumpConfig too.

This would make the implementation future-proof as well.

Prashansa-K:
> I can add a new method which has the new signature and make the old one an alias

This works too. Whatever is easy and not confusing can be done.

rspurgeon (approved):

PR #448, opened 2026-04-07
fix: validate explicit OIDC plugin config

mheap:
### Summary

Prevent `deck sync` from sending unsafe `openid-connect` configs that omit `cache_tokens_salt`. This fails fast after `_plugin_config` merging so decK does not pass `null` values through to Kong and accidentally trigger session-secret regeneration that invalidates active user sessions.

### Full changelog

* [Fix openid-connect plugin validation to require explicit non-empty `cache_tokens_salt` value]
* [Add unit tests covering `_plugin_config`-provided values and rejecting missing or `nil` fields]

### Issues resolved

https://github.com/Kong/deck/issues/1893

### Documentation

N/A

### Testing

- [x] Unit tests
- [ ] E2E tests


Prashansa-K:
@mheap I created a oidc plugin via admin api and UI (gw 3.14) and noticed that `cache_tokens_salt` is automatically generated but `session_secret` is set as null by default. Keeping the behaviour same, we should check only for `cache_tokens_salt` in deck. Otherwise a config dumped from UI would lead to a failure in deck diff/sync/validate.
Noticed the same with Konnect as well.

If Kong is not generating a `session_secret` by default, why would we enforce it via deck _only_?

I can do the remaining changes and test corrections here once the above query is clarified.

mheap:
@Prashansa-K I'm good with matching the UI. Thank you for investigating

Prashansa-K (approved):

Prashansa-K:
@shivaygupta-dotcom We need a new reviewer as I pushed some changes to correct the tests.

pkg/file/builder.go
```diff
@@ -18,7 +18,14 @@ import (
 	"github.com/kong/go-kong/kong"
 )
 
-const ratelimitingAdvancedPluginName = "rate-limiting-advanced"
+const (
+	ratelimitingAdvancedPluginName = "rate-limiting-advanced"
+	openIDConnectPluginName        = "openid-connect"
+)
+
+var openIDConnectRequiredConfigFields = []string{
```
shivaygupta-dotcom, line 26: In description, we have added `session_secret` also, but here we are just adding validation for `cache_tokens_salt`. Is it as per need or I am missing something ?

Prashansa-K: Ref: https://github.com/Kong/go-database-reconciler/pull/448#issuecomment-4258254652

Updated description.

shivaygupta-dotcom (approved):

PR #455, opened 2026-04-16
fix: preserve comment lines to hanlde big multi-line yamls

shivaygupta-dotcom:
### Summary

FTI: https://konghq.atlassian.net/browse/FTI-7356

We got above FTI where creating a request-termination plugin from the UI, then dumping it and running a diff on the same file, showed updated: 1 even though nothing had changed. This was caused by the long HTML body in the config exceeding YAML’s 80-character line limit, leading to line wrapping—especially for lines starting with CSS hex colors like #4c4c4c.

Fixed this by adding a check to correctly identify whether a line starting with # is an actual comment or just wrapped content, preventing false diffs and keeping the YAML intact.
### Full changelog

* [Fixed template rendering by adding a check to distinguish whether lines starting with # are actual comments or just wrapped content.]

### Issues resolve

Fix for : https://github.com/Kong/deck/issues/1958

### Documentation

### Testing

- [ ] Unit tests
- [ ] E2E tests


Prashansa-K:
@harshadixit12 Assigning this PR review to you as you may have better context here. You worked on a similar change here: https://github.com/Kong/go-database-reconciler/pull/351 - skipping template parsing for commented strings
Some changes in this PR modify the above PR's changes.

pkg/file/template.go
```diff
@@ -90,6 +91,8 @@ func indent(spaces int, v string) string {
 	return strings.ReplaceAll(v, "\n", "\n"+pad)
 }
 
+var commentRegex = regexp.MustCompile(`^\s*#($|\s|[^0-9a-fA-F])`)
```
harshadixit12: @shivaygupta-dotcom can you explain what this regex does, and how it solves the problem here?

pkg/file/template_test.go
```diff
@@ -157,3 +157,35 @@ func Test_renderTemplateSkipDoesNotRequireEnvVars(t *testing.T) {
 		t.Errorf("Unexpected error: %v", err)
 	}
 }
+
```
harshadixit12, line 162: Let's try adding one more test - where actual comment does not have whitespace after `#`
```
func Test_renderTemplateIgnoresCommentsNoSpace(t *testing.T) {
	content := `Hello, ${{ env "DECK_MY_VARIABLE" }}!
  #Also, ${{ env "DECK_NOT_SET_DOESNT_ERROR" }}!`

	expectedOutput := `Hello, my_value!`
	mode := EnvVarsExpand

	os.Setenv("DECK_MY_VARIABLE", "my_value")

	output, err := renderTemplate(content, mode)
	if err != nil {
		t.Errorf("Unexpected error: %v", err)
	}

	if output != expectedOutput {
		t.Errorf("Expected output %q, but got %q", expectedOutput, output)
	}
	os.Unsetenv("DECK_MY_VARIABLE")
}
```

harshadixit12:
@shivaygupta-dotcom two more questions - 
1. Which package / module / flow is responsible for the wrapping content in yaml within DecK or GDR?
2. How do we know what is an "actual comment" and what is wrapped content?

pkg/file/template.go
shivaygupta-dotcom: this helps in identifying if line starting with # is not a comment, if it has a space or a css hex colors.
but yeah as you provided below test, it will fail. I am figuring out more stable soln for this.

shivaygupta-dotcom:
> @shivaygupta-dotcom two more questions -
> 
> 1. Which package / module / flow is responsible for the wrapping content in yaml within DecK or GDR?
> 2. How do we know what is an "actual comment" and what is wrapped content?

1. By default yaml parsing is like this if one line is having more than 80 characters, yaml wraps it into another line, it is not related to decK or GDR. We currently have yaml/v2 and yaml/v3 in our configuration, we are facing this problem with both of them.
2. Regex which I have added is helping in identifying this but that is not enough as you mentioned.

pkg/file/template.go
```diff
@@ -115,14 +118,16 @@ func renderTemplate(content string, mode RenderEnvVarsMode) (string, error) {
...
 
-	// Parse content line by line, and ignore lines that start with #
 	var allContent bytes.Buffer
 	lines := strings.Split(content, "\n")
 	for i := 0; i < len(lines); i++ {
 		line := lines[i]
-		if !strings.HasPrefix(strings.TrimSpace(line), "#") {
-			allContent.WriteString(line + "\n")
+		if strings.HasPrefix(strings.TrimSpace(line), "#") {
+			// Remove any template expressions from the line to prevent accidental
```
harshadixit12: cc: @mheap for visibility - our original approach for ignoring env variable placeholders inside commented out lines fails inside multi-line strings in yaml.
Since we run templating before Unmarshalling, while processing line by line, it is not straightforward to know if a line starting with `#`  is a comment, or a multi line string
Instead of ignoring such lines, we are replacing the placeholder `${{env DECK_XYZ}}` with ""

This is not a perfect solution - while it will unblock FTIs like https://konghq.atlassian.net/browse/FTI-7356, it will fail in some rare cases - if a line has a css value starting with `#` and then a env variable placeholder - like `#4c4c4c; font-size: ${{env "DECK_RT_FONT_SIZE"}};` - Are we ok with this tradeoff?

mheap: Template expressions must be kept in the file. We can not strip them

harshadixit12: @shivaygupta-dotcom can you add examples of how the line containing css is being removed?

shivaygupta-dotcom: @mheap - 
Our current/original approach treated any YAML line starting with # as a comment and removed the entire line. While this worked for actual comments, it caused issues with large YAML payloads, especially those containing long HTML/CSS content.

When the YAML becomes large enough, it gets line-wrapped. In some cases, a wrapped line may begin with a CSS color value like #4cfcfc. Since our logic only checked for a leading #, such lines were incorrectly treated as comments and deleted entirely.

This led to silent data loss in the configuration and inconsistent diffs, particularly when entities were created via Admin API or UI. The FTI mentioned in this PR, along with [this](https://kongstrong.slack.com/archives/C03CTMSHP6C/p1776768120253669) issue, are a direct result of this behavior.

With this change, instead of removing any line that starts with #, we preserve the line and only strip environment variable expressions if they are present within it. Meaning instead of removing entire comment unlike our current/original approach, we are replacing placeholders like ${{env DECK_XYZ}} with "".

This prevents legitimate content (like CSS color codes) from being removed while still handling env expressions as intended.

There is a rare trade-off of this approach as mentioned by @harshadixit12  in his prev comment.

example: While executing below content, once the flow will reach to line starting with #116F9A, it will be removed entirely from the content.
```
email:
  - subject: Lion Rock
    htmlemail: <html><head><style>body { color: #4c4c4c; background-color: #ebeced; font-
      size: 14px; } .header { background-color: #006564; } .border { border-
      left: 8px solid #1B3668; border: 1px solid #BCBEC0; } .footer a { color: 
      #116F9A; }</style></head><body>Lion Rock Business Class Lounge Pass 
      cancellation.</body></html>
    textemail: ""
sms: ""
errors: null
```

harshadixit12: For visibility - we _might_ have another approach which identifies comments accurately, but slightly worse performance. Validating the correctness and performance.

mheap: Replacing placeholders with "" is not an acceptable fix. You _could_ replace it with something like `__NO_MATCH__ENV_DECK_XYZ__`, run the render then replace the old template strings if needed.

harshadixit12: Even in that case, an input line like `#4c4c4c; font-size: ${{env "DECK_RT_FONT_SIZE"}};` (which is not a comment) will fail to replace  ${{env "DECK_RT_FONT_SIZE"}} with the env var value - but yes, it will be more visible than replacing with ""

Another approach:
At the moment, we do templating on string([]bytes), and then unmarshal into `type Content`
Before we do templating, we _could_ unmarshal the []bytes -> `type Content` and back to []bytes - which would remove the comments - and then do templating. (There might be edge cases and performance drop, yet to test this thoroughly)

shivaygupta-dotcom: post discussing with @harshadixit12, we are moving ahead with approach suggested by @mheap.

pkg/file/template.go
```diff
@@ -115,22 +118,33 @@ func renderTemplate(content string, mode RenderEnvVarsMode) (string, error) {
...
+	counter := 0
 	var allContent bytes.Buffer
 	lines := strings.Split(content, "\n")
 	for i := 0; i < len(lines); i++ {
 		line := lines[i]
-		if !strings.HasPrefix(strings.TrimSpace(line), "#") {
-			allContent.WriteString(line + "\n")
+		if strings.HasPrefix(strings.TrimSpace(line), "#") {
+			line = templateExprPattern.ReplaceAllStringFunc(line, func(match string) string {
+				ph := fmt.Sprintf("__NO_MATCH__%d__", counter)
```
shivaygupta-dotcom, line 133: Using a numeric counter instead of string names avoids the additional overhead of name identification. Since this is an intermediate step in the flow, we only need to maintain a map and then replace the values back with the original environment variables later.

harshadixit12 (approved):

PR #473, opened 2026-05-26
fix: Added check in maskEnvVar func for multiple data types

### LINKED ISSUE #2047: deck generating invalid json output while masking field values.
While using deck with --json-output with a file that has env variables, deck is masking string values and integer values differently - with and without quotes.  Thus, the generated json is invalid. 

This also impacts kongctl apply command.

Ref: https://kongstrong.slack.com/archives/C07JBUNRW6N/p1778011971761239

shivaygupta-dotcom:
### Summary

Fixes broken `DECK_` environment variable value masking that produced corrupted and invalid output:

1. Invalid JSON with `--json-output`: Masked integer values were inserted without quotes (e.g., "retries": [masked]), producing unparseable JSON output that broke downstream tooling.
2. Over-aggressive substring replacement: Short env var values like `DECK_REDIS_DB`=0 caused every occurrence of that character across the entire output to be replaced — corrupting UUIDs, ports, and timeouts, and inflating output file sizes.

### Issues resolved

Fix: https://github.com/Kong/deck/issues/2047
https://github.com/Kong/deck/issues/2063

### Documentation

### Testing

- [x] Unit tests
- [ ] E2E tests

### DIFF
--- a/pkg/diff/diff_helpers.go
+++ b/pkg/diff/diff_helpers.go
@@ -211,21 +211,43 @@ var (
 )
 
 // MaskEnvVarValue masks DECK_ environment variable values in diff output.
-// Only exact, complete values are masked — never substrings within other values.
+// It identifies values in proper JSON/YAML positions (after `:` or as array elements)
+// and performs word-boundary substring replacement within those values. This prevents
+// corruption of non-value content (e.g., UUIDs in keys) while still masking secrets
+// that appear as part of larger strings.
 func MaskEnvVarValue(diffString string) string {
 	envVars := parseDeckEnvVars()
 	if len(envVars) == 0 {
 		return diffString
 	}
 
-	secrets := make(map[string]bool, len(envVars))
+	// Build sorted list of values (longest first) for substring replacement
+	var secrets []string
+	seen := make(map[string]bool, len(envVars))
 	for _, ev := range envVars {
-		if ev.Value != "" {
-			secrets[ev.Value] = true
+		if ev.Value != "" && !seen[ev.Value] {
+			secrets = append(secrets, ev.Value)
+			seen[ev.Value] = true
 		}
 	}
 	if len(secrets) == 0 {
 		return diffString
+	}
+	sort.Slice(secrets, func(i, j int) bool {
+		return len(secrets[i]) > len(secrets[j])
+	})
+
+	// Pre-compile word-boundary patterns once for all secrets
+	secretPatterns := make([]*regexp.Regexp, len(secrets))
+	for idx, secret := range secrets {
+		secretPatterns[idx] = regexp.MustCompile(`\b` + regexp.QuoteMeta(secret) + `\b`)
+	}
+
+	maskFn := func(s string) string {
+		for _, re := range secretPatterns {
+			s = re.ReplaceAllString(s, maskedValue)
+		}
+		return s
 	}
 
 	lines := strings.Split(diffString, "\n")
@@ -239,11 +261,12 @@ func MaskEnvVarValue(diffString string) string {
 			}
 			switch {
 			case sub[1] != "": // quoted string
-				if secrets[unescapeJSON(sub[1])] {
-					return match[:len(match)-len(`"`+sub[1]+`"`)] + `"` + maskedValue + `"`
+				masked := maskFn(sub[1])
+				if masked != sub[1] {
+					return match[:len(match)-len(`"`+sub[1]+`"`)] + `"` + masked + `"`
 				}
 			case sub[2] != "": // number
-				if secrets[sub[2]] {
+				if seen[sub[2]] {
 					prefix := match[:len(match)-len(sub[2])]
 					if isJSON {
 						return prefix + `"` + maskedValue + `"`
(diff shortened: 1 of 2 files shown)


pkg/diff/diff_helpers.go
```diff
@@ -190,9 +191,115 @@ func parseDeckEnvVars() []EnvVar {
...
+// that appear as part of larger strings.
 func MaskEnvVarValue(diffString string) string {
-	for _, envVar := range parseDeckEnvVars() {
-		diffString = strings.ReplaceAll(diffString, envVar.Value, "[masked]")
+	envVars := parseDeckEnvVars()
+	if len(envVars) == 0 {
+		return diffString
+	}
+
+	// Build sorted list of values (longest first) for substring replacement
```
Prashansa-K, line 226: Curious: Why are we sorting by longest first?

pkg/diff/diff_helpers.go
```diff
@@ -190,9 +191,115 @@ func parseDeckEnvVars() []EnvVar {
...
+	maskFn := func(s string) string {
+		for _, re := range secretPatterns {
+			s = re.ReplaceAllString(s, maskedValue)
+		}
+		return s
+	}
+
+	lines := strings.Split(diffString, "\n")
+	for i, line := range lines {
+		isJSON := strings.Contains(line, `":`)
```
Prashansa-K: 1. Why are we evaluating `isJSON` for every line? The entire diff would either be json or not, right?
2. `strings.Contains(diffString, `":`)` this seems like a weak approach to detect json.
Valid yaml lines can get past this. 
-> `example: 'format is "key": "value"'`
We don't know what sort of schema custom plugins use and what fields they may have. So, using a string comparison doesn't seem right.

pkg/diff/diff_helpers.go
```diff
@@ -190,9 +191,115 @@ func parseDeckEnvVars() []EnvVar {
...
+			}
+			switch {
+			case sub[1] != "": // quoted string
+				masked := maskFn(sub[1])
+				if masked != sub[1] {
+					return match[:len(match)-len(`"`+sub[1]+`"`)] + `"` + masked + `"`
+				}
+			case sub[2] != "": // number
+				if seen[sub[2]] {
+					prefix := match[:len(match)-len(sub[2])]
```
Prashansa-K: Prefix will always be a colon and an optional whitespace.
Do you think we need exact no. of whitespaces for numbers?
`port: 8080` or `port:     8080` is technically same for deck, right?

Prashansa-K: Also, this is just for "diffString"

pkg/diff/diff_helpers.go
```diff
@@ -190,9 +191,115 @@ func parseDeckEnvVars() []EnvVar {
...
+					if isJSON {
+						return prefix + `"` + maskedValue + `"`
+					}
+					return prefix + maskedValue
+				}
+			case sub[3] != "": // YAML unquoted
+				masked := maskFn(sub[3])
+				if masked != sub[3] {
+					return match[:len(match)-len(sub[3])] + masked
```
Prashansa-K: Same as L270.

Prashansa-K: We are mostly doing this extra work for getting colon and spaces between the colon and value right?

pkg/diff/diff_helpers.go
```diff
@@ -190,9 +191,115 @@ func parseDeckEnvVars() []EnvVar {
...
+		result := kvPattern.ReplaceAllStringFunc(line, func(match string) string {
+			sub := kvPattern.FindStringSubmatch(match)
+			if sub == nil {
+				return match
+			}
+			switch {
+			case sub[1] != "": // quoted string
+				masked := maskFn(sub[1])
+				if masked != sub[1] {
+					return match[:len(match)-len(`"`+sub[1]+`"`)] + `"` + masked + `"`
```
Prashansa-K, line 275: Same as L270.

pkg/diff/diff_helpers.go
shivaygupta-dotcom: We sort longest-first so that bigger secrets get replaced before smaller ones that might overlap. For example, if we have DECK_DB=redis and DECK_DB_HOST=redis.internal.example.com, and we replace redis first, the host becomes [masked].internal.example.com — leaking part of the secret. By doing the longer one first, we get [masked] cleanly, and then redis only matches other standalone occurrences.

pkg/diff/diff_helpers.go
shivaygupta-dotcom: The prefix calculation (match[:len(match)-len(sub[2])]) simply preserves whatever was before the number — which is the colon and any whitespace (e.g., : or :\t). Then based on isJSON, we decide whether to output "[masked]" (with quotes, to keep JSON valid) or just [masked] (without quotes, for YAML/plain text).

So the entire purpose of computing prefix is just: keep the original : separator intact, and only replace the value portion with the appropriate masked format.

Prashansa-K: So, why can't we hardcode `:` instead of computing it?

shivaygupta-dotcom: so, for numbers and YAML it's always : so hardcoding would work there. But the quoted string case uses :\s*" which matches both :"val" (compact JSON) and : "val" (pretty JSON), so we need to preserve whatever spacing was actually there. 
I just kept all three cases consistent rather than mixing approaches — it's a zero-cost string slice anyway, and saves us from a subtle bug if formatting ever changes.

Prashansa-K: I would suggest to hardcode where it's possible and add a comment about the pretty json for case sub[2].

shivaygupta-dotcom: done.

Prashansa-K (approved):

PR #500, opened 2026-07-13
fix: handle nil and empty Headers map equivalence in Route equality checks

### LINKED ISSUE #2147: Empty headers causing false update alerts
deck keep updating a route object when we set `headers={}` even though no any changes

VarunAthreya:
### Summary

Fix Route equality comparison so nil and empty Headers maps are treated as equivalent, preventing Deck from reporting unnecessary updates when comparing the desired state (empty headers from YAML) with Kong's response (nil headers).

### Full changelog

* [Fix] Handle nil and empty Headers map equivalence in Route equality checks
* [Test] Add TestRouteEqual_HeadersNilVsEmpty to verify nil vs empty headers comparison

### Issues resolved

FIx: https://github.com/Kong/deck/issues/2147

### Testing

- [x] Unit tests
- [ ] E2E tests


pkg/state/types.go
```diff
@@ -182,6 +182,17 @@ func (r1 *Route) EqualWithOpts(r2 *Route, ignoreID,
 	sort.Slice(r1Copy.Tags, func(i, j int) bool { return *(r1Copy.Tags[i]) < *(r1Copy.Tags[j]) })
 	sort.Slice(r2Copy.Tags, func(i, j int) bool { return *(r2Copy.Tags[i]) < *(r2Copy.Tags[j]) })
 
+	// A nil Headers map and an empty (non-nil) one are semantically identical,
+	// but reflect.DeepEqual below treats them as different, causing deck to
+	// report a spurious update every time (e.g. "headers: {}" in the desired
+	// state vs. Headers == nil as returned by Kong).
+	if len(r1Copy.Headers) == 0 {
```
Prashansa-K, line 189: This looks good. 
Could you confirm from the gateway team if we need to consider headers as order-sensitive or not? If they aren't order-sensitive, let's also add sorting for them, like we have for tags above.

VarunAthreya:
```go
Headers       map[string][]string `json:"headers,omitempty" yaml:"headers,omitempty"`
``` 

[go-kong/kong/route.go](https://github.com/Kong/go-kong/blob/dae27d2546f4f0e9381ba8aff87f885569e14fdf/kong/route.go#L10)

the headers type is a map rather than a slice, so I don't think we need to sort it in this case.

Prashansa-K (approved):

PR #506, opened 2026-08-04
fix: normalize empty arrays to nil for consistent comparison in config tree

### LINKED ISSUE #2175: decK reports perpetual diff for plugin config fields with empty array ([]) values
`decK diff` / `decK gateway sync` reports a perpetual diff for any plugin config field whose value is an empty array (`[]`). The diff never converges, running sync/diff repeatedly against the *same* unchanged source config continues to show a change for that field on every run, instead of reaching a clean "no differences" state.

This is **not specific to a single plugin**. Any plugin config schema that includes an array-typed field will reproduce the issue as long as that field's value is `[]`.

VarunAthreya:
### Summary

Fixed a perpetual diff where `deck diff` / `deck sync `never converged for any plugin config field set to an empty array (`[]`). One side of the comparison held `[]` while the other held nil/absent, so reflect.DeepEqual always reported a change and an update was queued on every run. Empty arrays are now normalized to `nil` at every depth of the config tree before comparison, so `[]` and an unset field compare equal and the diff reaches a clean `"no changes" `state.

### Full changelog

* [Fix] Perpetual diff issue for plugin config when an empty array is passed.

### Issues resolved

Fix: https://github.com/Kong/deck/issues/2175

### Testing

- [x] Unit tests
- [ ] E2E tests

### DIFF
--- a/pkg/state/types.go
+++ b/pkg/state/types.go
@@ -721,7 +721,6 @@ func sortNestedArraysBasedOnSchema(m map[string]any, schema gjson.Result) map[st
 	for k, v := range m {
 		switch value := v.(type) {
 		case []any:
-			// Normalize empty arrays to nil so an explicit `[]`
 			// Normalize empty arrays to nil so an explicit `[]` and an
 			// explicit nil value for the same key compare equal at every
 			// depth of the config tree. Note this does not equate `[]` with
--- a/pkg/state/types.go
+++ b/pkg/state/types.go
@@ -722,7 +722,11 @@ func sortNestedArraysBasedOnSchema(m map[string]any, schema gjson.Result) map[st
 		switch value := v.(type) {
 		case []any:
 			// Normalize empty arrays to nil so an explicit `[]`
-			// and an absent/nil field compare equal at every depth of the config tree.
+			// Normalize empty arrays to nil so an explicit `[]` and an
+			// explicit nil value for the same key compare equal at every
+			// depth of the config tree. Note this does not equate `[]` with
+			// an absent key: a key that isn't present is never added to the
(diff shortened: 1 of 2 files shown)


pkg/state/types.go
```diff
@@ -721,6 +721,13 @@ func sortNestedArraysBasedOnSchema(m map[string]any, schema gjson.Result) map[st
 	for k, v := range m {
 		switch value := v.(type) {
 		case []any:
+			// Normalize empty arrays to nil so an explicit `[]`
+			// and an absent/nil field compare equal at every depth of the config tree.
```
Prashansa-K: Implementation looks good to me.
However, the comment is misleading. "an absent field" is not taken into consideration here. The tests say the same thing.
So, let's remove "absent" from the comment to avoid confusions in future.


```suggestion
			// Normalize empty arrays to nil so an explicit `[]` and an
			// explicit nil value for the same key compare equal at every
			// depth of the config tree. Note this does not equate `[]` with
			// an absent key: a key that isn't present is never added to the
			// map, so it stays distinct from an explicit empty array/nil.
```

pkg/state/types.go
```diff
@@ -721,6 +721,17 @@ func sortNestedArraysBasedOnSchema(m map[string]any, schema gjson.Result) map[st
 	for k, v := range m {
 		switch value := v.(type) {
 		case []any:
+			// Normalize empty arrays to nil so an explicit `[]`
```
Prashansa-K: Why is this line repeated? 🤔

VarunAthreya: my bad, removed it

Prashansa-K (approved):

PR #507, opened 2026-08-06
fix: preserve explicit nil config values when skipping schema defaults

### LINKED ISSUE #1824: A field with non-null default, set as null by enduser is getting omitted in dump while using --skip-defaults.
- configure the 'session_redis_host' field with value as empty ('null') in the openid-connect plugin
- run 'deck gateway dump --skip-default' command to dump the configurartion;
- run 'deck gateway diff xxxx.yaml'

equistar22:
### Summary

This PR fixes the comparison so an explicit `null` is checked against the schema default like any other value, and kept in the output when it differs.

### Full changelog

* [Fix] `--skip-defaults` no longer drops plugin config fields explicitly set to
  `null` by the user when the field's schema default is non-null.

### Issues resolved

Fixes : https://github.com/Kong/deck/issues/1824

### Documentation

### Testing

- [x] Unit tests
- [ ] E2E tests

### DIFF
--- a/pkg/dump/skip_defaults.go
+++ b/pkg/dump/skip_defaults.go
@@ -523,12 +523,21 @@ func compareMaps(fieldMap, defaultMap reflect.Value) any {
 			continue // Skip unexported fields
 		}
 
-		fieldVal = reflect.ValueOf(fieldVal.Interface())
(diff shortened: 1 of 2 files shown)


equistar22:
Cancelling this PR as we need to do several changes at gateway level.

PR #509, opened 2026-08-13
Fix/env var masking secrets

shivaygupta-dotcom:
### Summary

SUMMARY_GOES_HERE

### Full changelog

* [Implement ...]
* [Fix ...]

### Issues resolved

Fix #XXX

### Documentation

### Testing

- [ ] Unit tests
- [ ] E2E tests


pkg/file/types.go
```diff
@@ -713,13 +713,13 @@ func (p FPlugin) sortKey() string {
 type FConsumer struct {
 	kong.Consumer `yaml:",inline,omitempty"`
 	Plugins       []*FPlugin               `json:"plugins,omitempty" yaml:",omitempty"`
-	KeyAuths      []*kong.KeyAuth          `json:"keyauth_credentials,omitempty" yaml:"keyauth_credentials,omitempty"`
-	HMACAuths     []*kong.HMACAuth         `json:"hmacauth_credentials,omitempty" yaml:"hmacauth_credentials,omitempty"`
-	JWTAuths      []*kong.JWTAuth          `json:"jwt_secrets,omitempty" yaml:"jwt_secrets,omitempty"`
-	BasicAuths    []*kong.BasicAuth        `json:"basicauth_credentials,omitempty" yaml:"basicauth_credentials,omitempty"`
-	Oauth2Creds   []*kong.Oauth2Credential `json:"oauth2_credentials,omitempty" yaml:"oauth2_credentials,omitempty"`
+	KeyAuths      []*kong.KeyAuth          `json:"keyauth_credentials,omitempty" yaml:"keyauth_credentials,omitempty" credential:"keyauth_credential,key"`            //nolint
```
harshadixit12, line 716: Why do we need the `credential` tag here? where / how are we using this custom tag?

pkg/file/readfile.go
```diff
@@ -58,6 +58,29 @@ func getContent(filenames []string, mode RenderEnvVarsMode) (*Content, error) {
 	return &res, nil
 }
 
+func getMockContent(filenames []string, mode RenderEnvVarsMode) (string, error) {
```
harshadixit12, line 61: I'd name this differently - `getMockContent` doesn't explain what this function is supposed to do.

harshadixit12: Also, let's add a comment on top explaining what the function does.

pkg/file/readfile.go
```diff
@@ -136,6 +159,18 @@ func readContent(reader io.Reader, mode RenderEnvVarsMode) (*Content, error) {
 	return &result, nil
 }
 
+func readSkipContent(reader io.Reader, mode RenderEnvVarsMode) (string, error) {
```
harshadixit12, line 162: Same here - `readSkipContent` doesn't explain what is being done - is this function meant for not replacing placeholders with env vars during templating?

pkg/file/readfile.go
```diff
@@ -58,6 +58,29 @@ func getContent(filenames []string, mode RenderEnvVarsMode) (*Content, error) {
...
+		if err != nil {
+			return "", err
+		}
+		for filename, r := range readers {
+			content, err := readSkipContent(r, mode)
+			if err != nil {
+				errs = append(errs, fmt.Errorf("reading file %s: %w", filename, err))
+				continue
+			}
+			res = content
```
harshadixit12, line 75: Does this implementation work when we pass multiple files? I think no - because we aren't merging the result - instead overwriting it. (verify & refer to `getContent()`)

pkg/file/readfile.go
harshadixit12: Note - `getContent()` returns type `Content` - but `getMockContent` returns string - intentional?

pkg/file/reader.go
```diff
@@ -60,6 +60,14 @@ func GetContentFromFilesWithEnvVars(filenames []string, mode RenderEnvVarsMode)
 	return getContent(filenames, mode)
 }
 
+func GetMockContentFromFiles(filenames []string, mode RenderEnvVarsMode) (string, error) {
```
harshadixit12, line 63: Same here - intended that returning string instead of type `Content`?

harshadixit12, line 63: IMO we should add unit tests in reader - to read file, directory, and multiple files - for both existing functions in use, and the new functions being added.

pkg/diff/diff.go
```diff
@@ -615,33 +629,91 @@ type Stats struct {
 
 // generateDiffStringWithCache is like prev generateDiffString but uses a precomputed
```
harshadixit12: This comment is displaced

pkg/diff/mask_by_fieldname.go
```diff
@@ -0,0 +1,334 @@
...
+var maskedValueChanged = maskedValue + string(rune(0x200D))
+
+// maskEntityPairByFieldNames masks the old and new sides of an entity
+// TOGETHER, field name by field name, so it can tell whether the real
+// (pre-mask) value actually changed. Unmodified secret fields are masked
+// identically on both sides; changed ones get a visually-identical but
+// byte-different placeholder on the new side, preserving the "this changed"
+// signal without revealing what changed.
+// Neither original object is mutated.
+func maskEntityPairByFieldNames(oldEntity, newEntity any, secretFields map[string]bool) (any, any) {
```
harshadixit12, line 27: I see the return type is `any` - while that is acceptable in some cases, it is not idiomatic go convention to return errors as `any` - let's return errors separately.
It also makes the code more readable, and bugs less likely.
We shouldn't need functions like `isErrorValue()` at all

pkg/diff/diff.go
```diff
@@ -615,33 +629,91 @@ type Stats struct {
...
 // environment variable cache to avoid redundant masking computations.
+// isErrorValue checks if a value is an error (used to detect masking failures)
+func isErrorValue(v any) bool {
```
harshadixit12, line 629: We shouldn't need a function like this - explained in this comment - https://github.com/Kong/go-database-reconciler/pull/509/changes#r3794159214

harshadixit12 (commented):
Partial review - I've not yet reviewed any core implementation of secret mapping, or masking.

